PR #56, opened 2019-01-09
pagination should not show when zero rows

jbetancur:
* hide pagination when zero rows


ajai06:
is it possible to show pagination when zero rows ?

PR #100, opened 2019-06-12
Update README.md

hiukky:
double .. : )


jbetancur:
Thank you for taking the time to submit a PR!

So, I would prefer to not change the API for component props. What I was planning on doing was creating a new prop `paginationComponentOptions` that you can pass an object with any number of options. For the built in pagination component i would have something like `{ rowsPerPageText: ..., etc...}` and that can be overridden: e.g.

```
paginationComponentOptions={ rowsPerPageText: 'Custom Rows Per Page' }
```

This also allows those that want to develop a custom component to have a hook to pass in their own customization props.

hiukky:
@jbetancur Yeah .. yesterday I did passing the prop straight away .. but today I had the same idea of ​​creating a prop with options .. this really is good!
I do not know if you looked at my fork .. I'm trying to

Issue #99, opened 2019-06-11
Issue with setting Page number


rakeshmanjunatha:
In a table by default page number will be 1 but i want to set it to any number dynamically.
Any idea?

Thanks

jbetancur:
I agree that this should be a feature. I’ll see if I can get this feature done this week

rakeshmanjunatha:
Thanks for the update, but i'm trying to update this package in my system, its not happening getting older one only.

jbetancur:
Hi,

That’s because I didn’t update anything yet ;)

rakeshmanjunatha:
okay fine, it will be helpful if you push the latest one.

jbetancur:
Fixed in https://github.com/jbetancur/react-data-table-component/pull/102

I'll cut a release this weekend once I get your other issue #101 resolved

rakeshmanjunatha:
Hi,

Still its not working, when i'm trying to set the page number dynamically. but its working fine with initial default page number.

Actually i'm storing the page number in URL on onChangePage event.

Scenario: On page load if the URL has "page" number parameter, i'm trying to assigning value to **paginationDefaultPage**  using setState but its not working.

PR #102, opened 2019-06-12
Add paginationDefaultPage prop to DataTable for setting the initial page number

Lknechtli:
Add a `defaultPage` prop to DataTable.
This is useful when you want to store the page size / number in a url parameter and allow navigating directly to a page in the table.

I think this is what #99 is asking for, but either way it would be very useful to have.


jbetancur:
Thank you for the PR! I'll have a look at this a little later today.

jbetancur:
Ok, just a few things and I can get this merged in:

* To keep the naming convention can you rename to `paginationDefaultPage`?
* Make sure the tests pass `yarn test`
* Update the README with the new property

Lknechtli:
:+1: I'll have some time in the morning to do it.

Lknechtli:
image
@jbetancur Pushed the fixes

allencazar:
@jbetancur , is there a proper way to control or set the current page?
- wanted to go to last page or any page programmatically. 
- highlight last/new added row.

PR #104, opened 2019-06-14
Add custom texts to pagination component

metabaronespinosa:
Just added the possibility to pass custom texts to pagination component

### DIFF
--- a/src/DataTable/DataTable.js
+++ b/src/DataTable/DataTable.js
@@ -237,6 +237,8 @@ class DataTable extends Component {
       subHeaderAlign,
       subHeaderWrap,
       subHeaderComponent,
+      rowsPerPageText,
+      ofText
     } = this.props;
 
     const {
@@ -318,6 +320,8 @@ class DataTable extends Component {
                     currentPage={currentPage}
                     rowsPerPage={rowsPerPage}
                     theme={theme}
+                    rowsPerPageText={rowsPerPageText}
+                    ofText={ofText}
                   />
                 </TableFooter>
               )}
--- a/src/DataTable/Pagination.js
+++ b/src/DataTable/Pagination.js
@@ -54,6 +54,8 @@ export default class Pagination extends PureComponent {
     onChangeRowsPerPage: PropTypes.func.isRequired,
     theme: PropTypes.object.isRequired,
     currentPage: PropTypes.number.isRequired,
+    rowsPerPageText: PropTypes.string,
+    ofText: PropTypes.string,
   };
 
   getNumberOfPages() {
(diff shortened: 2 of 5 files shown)


hiukky:
@metabaronespinosa your commit reminds me of my ..  : )

Lknechtli:
I think it might be better just to add a localizationText prop. Passing a separate prop for each string seems slightly silly when you consider that most applications that support localization store the strings in a separate file.
Doing something like this seems a bit silly:
```javascript
import L10nJson from "L10n.json";
const { rowsPerPageText, ofText, ...otherTextFieldsAsProps} = L10nJson;
const LocalizedDataTable= (...props) => <DataTable {...props} rowsPerPageText={rowsPerPageText} ofText={ofText} {...otherTextFieldsAsProps} />;
```
when you can just do
```javascript
import L10nJson from "L10n.json";
const LocalizedDataTable= (...props) => <DataTable {...props} localizationText={L10nJson} />;
```

hiukky:
@Lknechtli yea!  I had done this in a first case, but I was looking for another way to improve it!  Your idea is cool ..

jbetancur:
I was planning on adding a new prop `paginationComponentOptions` that takes an object of values that can be used to override the default pagination component as well as used for custom component development when you want to provide a hook.

e.g.
```
onChangeRowsPerPage={ rowsPerPageText: 'Rows per Page' }
```

hiukky:
@jbetancur 
Very good! here are but a few options we need hehe! : )
image

jbetancur:
@hiukky 1.5.4 now has `paginationComponentOptions`. As for progress, noData and expander you can just use: progressComponent, noDataComponent, and expandableRowsComponent respectively. These props take either a string or component:

```
class MyComponent extends Component {
  render() { 
    return (
      <DataTable
        title="Arnold Movies"
        columns={columns}
        data={data}
        progressComponent="Un Momento"
        noDataComponent="Nada"
        expandableRowsComponent="actually, you would always use a custom component here"
        pagination
        paginationComponentOptions={{ rowsPerPageText: 'Filas por pagina', rangeSeparatorText: 'de' }}
      />
    )
  }
);
```

hiukky:
@jbetancur 
Good in both cases of progressComponent and expandableRowsComponent will not go right because here 
https://github.com/jbetancur/react-data-table-component/blob/master/src/DataTable/propTypes.js#L101-L103 
you are passing an element, then the user would also have to pass an element and if it does not do this an error will be triggered against the defined PropTypes! what I had done was left just the string there and put the elements in the main file rendering them ... and also ends up not having a standardization since some not very experienced user may think that if you put progressText as in `paginationComponentOptions = {rangeSeparatorText : '....'} `will work do not you think?

I think defining a prop getting all the string used would be more interesting than dividing by categories .. pagination ..body ..head `DataTableText = {prop: '...', prop2: '...'}` For example

jbetancur:
You are correct about the prop type not accepting string for expandableComponet, but Expandable rows does not need translation or options as it is literally meant for you to pass in a custom component 100% of the time.

progressComponent accepts a string however, here I would expect someone to pass a progress component but you can pass text just the same. 

For pagination I think as long as the api is documented clearly - which it is on the readme and storyboard and we have examples it should work. I’m not a fan of managing language in the library, but for pagination I provided the hooks to do so since it’s a bit of an outlier. 

Hope that helps but 1.5.5 which I released today has the paginationOptions for you to use

metabaronespinosa:
Hey, thank you very much for the last release, it totally solves what I wanted to reach

dinizime:
Sorry for re-opening. But how about localization in Selectable Rows? (... item selected)

jbetancur:
@dinizime I was thinking about this the other day. I'll open a separate issue for this, though I don't think it will be as easy as adding localization. Need to re-thing the context menu a bit

PR #111, opened 2019-06-16
Align columns by expanding rows

hiukky:
This was bothering..

image

Lined up

image


jbetancur:
Unfortunately, this doesn't work when you use `selectableRows` as well. I think the problem lies in the padding logic for Cell and TableCol.

jbetancur:
@hiukky I appreciate the PR. The issue was a bit more involved though. I will merge a similar fix to yours into a 1.6.1 release. Thanks again!

https://github.com/jbetancur/react-data-table-component/pull/116

PR #122, opened 2019-06-29
[Feature] Disable expander in some rows

### LINKED ISSUE #112: Custom expander button and disable expandable feature for specific rows
Hi,

I'm looking for adding a custom element instead of the default expander button, is there any way to do that with the current props of the component?

And also, I want to disable the expandable feature for some specific rows due to data validations, but I don't know if this is possible.

I hope you can help me clarify these questions.

Thanks!

Wiland:
Add the feature to disable row expander only for a few rows. I asked you for this some days ago, and I got enough time to do some coding and make it works.

I did some stuff based on your [response](https://github.com/jbetancur/react-data-table-component/issues/112#issuecomment-503649660) to my question...
> As for disabling the expander for specific rows, this is not currently possible. I imagine this feature could just be a prop that you can set on your data that the table will look for and disable the expander.
> 
> ```
> const myData = [{ id: 1, name: 'disableMe', disabled: true, ... }]
> ```
> 
> ```
> expandable
> expanderDisabledProp="disabled"
> ```
> 
> Then you just manage that prop however you map your data set.
> 
> I don't have time to add this right now, but I'm open to PR requests ;)

Let me know if you have any recommendation.
Fixes #112

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ Nothing new here - we are using an array of object literals and properties to de
 | selectableRowsComponent | func | no |  | Override the default checkbox component - must be passed as a function (e.g. `Checkbox` not `<Checkbox />`) |
 | selectableRowsComponentProps | object | no |  | Additional props you want to pass to `selectableRowsComponent`. See [Advanced Selectable Component Options](#advanced-selectable-component-options) to learn how you can override indeterminate state |
 | expandableRows | bool | no | false | Whether to make a row expandable, if true it requires an `expandableRowsComponent`. It is **highly recommended** your data set have a unique identifier defined as the `keyField` for row expansion to work properly.
-| expanderDisabledField | string | no | | React Data Table looks for this property for each item in your data to check if that item can be expanded or not. You must set a bool value in the `expanderDisabledField` of your data if you want to use this feature.
+| expandableDisabledField | string | no |  | React Data Table looks for this property for each item in your data to check if that item can be expanded or not. You must set a bool value in the `expandableDisabledField` of your data if you want to use this feature.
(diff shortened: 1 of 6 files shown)


jbetancur:
Nice! And thank you for taking the time to submit a PR! 

I'll have a look over the weekend and test, but a couple of things that stand out are that we should keep the expander props consistent. `expanderDisabledField ` should be `expandableDisabledField`. Also, I think the defaultProps should `disabled` instead of `''`

Wiland:
Hi @jbetancur, I fixed the prop name as you recommend.

About the default prop, I think that it should keep as '' in order to avoid any unexpected behavior because it is possible to have a data field named 'disabled' but not indicating that the item shouldn't be expandable. 

That's my thought, please let me know if you agree.

jbetancur:
That's a good point. Keep it as `''`

jbetancur:
Ok just got a chance to test and it seems that you cannot tell when an expander is actually disabled. The fix would be to add  a`&:disable...` selector to the ExpanderButton` and override the color, however, it seems I made the expander icons css background url's which you cannot override.

So... I need to re-think this and move the expander button into a table Prop e.g. `expandableIconExpanded` and expandableIconCollapsed` or something like that. 

Once I fix that you can rebase and we can merge this PR.

jbetancur:
@Wiland I've just merged in https://github.com/jbetancur/react-data-table-component/pull/124

It was a big change and as I expect you will have to rebase your PR (sorry!).  As I mentioned before this the expander Icon is now something that can be passed in as a component instead of being set as a css backgroud url, which means when applying styles such as `disabled` should now work properly

PR #148, opened 2019-07-13
Feature/row expanded by default

Wiland:
Hi @jbetancur,

This PR is related to a new feature that I need in a project I'm working on. It's the possibility of setting a field in each data item (similar to the expandableDisabledField) that contains a boolean to render the row expanded by default on the initial render.

I also added a rule exception in the eslint config file in order to avoid an error in Windows OS related to linebreaks.

Let me know if there's anything that I need to fix before merging this PR.

Thanks!


src/DataTable/__tests__/DataTable.test.js
```diff
@@ -457,6 +457,22 @@ describe('DataTable::expandableRows', () => {
...
   });
+
+  test('should render correctly when defaultExpandedField is true', () => {
+    const mock = dataMock();
+    mock.data[0].defaultExpanded = true;
+    const { container } = render(
+      <DataTable
+        data={mock.data}
+        columns={mock.columns}
+        defaultSortField="some.name"
```
jbetancur: minor issue: I would remove this from the test since you are not setting a sortable column.

src/DataTable/TableRow.js
```diff
@@ -70,6 +70,16 @@ class TableRow extends PureComponent {
     expanded: false,
   }
 
+  componentDidMount() {
+    const { row } = this.props;
+    const { defaultExpandedField } = this.context;
+    if (defaultExpandedField) {
```
jbetancur: super minor nitpick ;) put a line a break between if statements and variables declarations

Wiland: Done

src/DataTable/__tests__/DataTable.test.js
Wiland: You're totally right, I just removed it from the test code

src/DataTable/TableRow.js
```diff
@@ -70,6 +70,16 @@ class TableRow extends PureComponent {
     expanded: false,
   }
 
+  componentDidMount() {
```
jbetancur: I just noticed this but I would suggest not calling setState on componentDidMount. Instead, just set a default state in the constructor.

I think you could even just change the this.state to

```
  state = {
    expanded: row[defaultExpandedField] || false,
  }
```

Wiland:
@jbetancur, I solve both of the comments that you check and I ran all the tests on my local machine and all works... But in the CI is not working, can you help me check this?

jbetancur:
@Wiland I made several changes this weekend so you will need to rebase to get that failing snapshot to pass.

Also see my last comment regarding calling setState in componentDidMount

src/DataTable/TableRow.js
Wiland: The thing here is that defaultExpandedField comes from context, not props... And in the constructor method, context isn't available yet. That's the real reason for using setState in the componentDidMount.

But, if you know a better way to do this, please tell me how we can do it.

jbetancur: I believe you can just do this.context. if that doesn't work just set things up in a `constructor`

```
  state = {
    expanded: this.props.row[this.context.defaultExpandedField] || false,
  }
```

jbetancur: Or if this.state doesn't work for some reason:
```
constructor(props) {
  super (props);
  this.state = {
    expanded:props.row[this.context.defaultExpandedField] || false,
  }
}
```

jbetancur: Ahh ok. I see. That's interesting that this.context is not available on the constructor

Wiland: I already tried that, but this.context in the constructor is undefined

jbetancur: so you can access context as a parm on the constructur
```
  constructor(props, context) {
    super(props, context);

    this.state = {
      expanded: props.row[context.defaultExpandedField],
    };
  }
```

Wiland: What I think I can do is this in the renderRows method:

`this.calculateRows().map((row, i) => (
<TableRow
          key={row[keyField] || i}
          row={row}
          defaultExpanded={row[defaultExpandedField] || false}
          onRowClicked={this.handleRowClicked}
          onRowSelected={this.handleRowSelected}
        />
      ))`

Wiland: @jbetancur I did the change from my [last comment](https://github.com/jbetancur/react-data-table-component/pull/148#discussion_r303257228). I think this a better solution to avoid using setState in the componentDidMount. Please check it and tell me if it's okay.

Wiland: > so you can access context as a parm on the constructor
> 
> ```
>   constructor(props, context) {
>     super(props, context);
> 
>     this.state = {
>       expanded: props.row[context.defaultExpandedField],
>     };
>   }
> ```

I made a little digging and I found that this context param in the constructor method seems to be (or going to be) deprecated. Look at this [issue](https://github.com/facebook/react/issues/6598)

jbetancur: Bummer! Although it feels weird to have to pass in props that are also available on the context.  But that is a design issue with TableRow; which doesn't really need access to context since it is 1 level deep. I'll fix that in the future

jbetancur:
I need to do a couple more commits, but I should get a 2.1.0 release out shortly

PR #156, opened 2019-07-16
disable checkboxes for seletec rows

elshahat:
Add the option to disable checkboxes for selected rows by adding a new prop called "checkboxStatusSelector" where the value passed to this prop is the key in the object passed to the table and the value for this key in the object should be a boolean.

If the value is true the checkbox will not be disabled and if it is false the checkbox will be disabled

We can agree on adding some more changes like selecting the proper prop name and the the const names.


jbetancur:
@elshahat It's been a busy week so I have not had a chance to review. Once I get a chance I'll provide some commentary.

P.S. I am working on some internal architectural changes that may conflict with this PR.

jbetancur:
@elshahat a few things:

* There is no need to include dist or omit it from `.gitignore`. RDT uses `npm publish` which takes care of generating the artifacts.
* You should have tests cases added for DataTable.
* The prop name `checkboxStatusSelector` should probably be `selectableRowsDisabledField` so it is better categorized. (as a side note, in a future version I may look to collapsing feature props into a single prop object . e.g. `selectableOptions<Object>`

jbetancur:
superseded by #210

David-DB88:
Hi, i am using  react-data-table-component, i have a question, please tell me how i can disabled some checkboxes if i want, do you have any example? show me please it would be very help.

PR #158, opened 2019-07-21
pass rowIndex argument to cell function

manutdkid77:
It would be nice if we had access to the index of the row being rendered in the `cell` function. 

### API Change
`cell` function can now accept an additional argument, which indicates the index of the row.
`cell: row,rowIndex=>{ //doSomething }`

### Intended Use Case
For example, if we want to display the serial number column of each item in the `data` array
```
{
    name: 'Serial No',
    cell: (row,rowIndex) => rowIndex,
}
```

Please let me know if any changes are needed ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ Nothing new here - we are using an array of object literals and properties to de
 | selector | string | yes      | the propery in the data set e.g.  `property1.nested1.nested2`.                                                |
 | sortable | bool   | no       | if the column is sortable                                                                                     |
 | format   | func   | no       | format the selector e.g. `row => moment(row.timestamp).format('lll')`                                         |
-| cell     | func   | no       | for ultimate control use `cell` to render your own custom component! e.g `row => <h2>{row.title}</h2>`  **Negates  `format`** |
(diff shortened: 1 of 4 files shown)


jbetancur:
Thanks for the PR! 

Why do you need access to the index? Each row should have an `id` property or one that you can override with `keyField` if you need to access a particular row in your dataset

manutdkid77:
Thanks for reviewing the PR John 🙂

Sometimes an API returns a list of users (or some data) where in the ID column may not be in a sequence starting from 1,2,3,4,5.... 
Having access to rowIndex will help to create a table as shown below

| Serial No (created using rowIndex) | PersonID | Name   |
|-----------|----------|--------|
| 1         | 136      | John   |
| 2         | 231      | Jane   |
| 3         | 432      | Jim    |
| 4         | 12       | James  |
| 5         | 3        | Joanna |

jbetancur:
It seems like you want to use the index field for sorting/order? If so, can you just map an index into your dataset before passing it to DataTable?
e.g.
```
const dataWithSerials = data.map((data, index) => ({ serial: index, ...data })); 
```

manutdkid77:
Yes, using `map` to create an index field in my dataset is a better idea. Thanks John

Issue #154, opened 2019-07-15
Migrate to Hooks


jbetancur:

jbetancur:
#162 Just testing and doing some performance checks to make sure all is OK

PR #164, opened 2019-07-27
migrate components to react hooks

jbetancur:


jbetancur:
This is a rather bit commit, but there were extensive architectural changes that were required. The key benefit is that Table performance is somewhere around 50% faster than before optimizations

jbetancur:
* Closes #154 
* Move to React Hooks
* Improves tale performance by around 50% by reducing cases where the table was re-rendering more than necessary

PR #167, opened 2019-07-29
Reset pagination on new column sort

JeremyEllingham:
Updated so that column sorts reset Pagination to first page.  I am happy to make this an optional prop, but I don't see a need, this is the default behaviour a user will always expect anyway (it doesn't really make sense to bounce between page 2 and 9, or 3 and 7 when doing alternating 'asc'/'desc' sorts, rather if the user sorts by a given column then they clearly intend to be back at page 1 of the data set with that new sort applied).

Bonus: Updated documentation to make contribution setup a little clearer particularly the required Node version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -676,7 +676,15 @@ If you would like to customize the layout components of React Data Table using s
 # Development
 
 ## Setup
-Install the latest [Node JS LTS](https://nodejs.org/) and [Yarn](https://yarnpkg.com) and simply run `yarn` or `yarn install` command in the root and stories directory.
+
+Make sure your local Node version is set to `10.15.0` using [nvm](https://github.com/nvm-sh/nvm) or [n](https://github.com/tj/n).  This specific version is required, as Node v12+ and potentially other versions have some broken dependencies with Storybook.
+
+```bash
+$ node --version
+# v10.15.0
+$ yarn install
+$ yarn storybook # Local hotreloading example component available at http://localhost:6006
+```
 
 > It is advised to run the script whenever NPM packages are installed.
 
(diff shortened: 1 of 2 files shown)


jbetancur:
So we can change the current Page when sorting to 1 by just changing the state in the reducer

```
    case 'SORT_CHANGE': {
      const { sortColumn, sortDirection } = action;

      return {
        ...state,
        sortColumn,
        sortDirection,
        currentPage: 1,
      };
    }
```

Then in DataTable we need to just wrap it in a useEffect. The only problem is that it will trigger onChangePage on each page load. I'll write something like useComponentDidUpdate hook to fix this, but this is what the effect would look like

```
  useEffect(() => {
    onChangePage(currentPage, totalRows);
  }, [currentPage, onChangePage, totalRows]);
```

jbetancur:
Let me know if you want to try making those changes, In the meantime, I am going to push up a new hook to deal with these callbacks.

jbetancur:
Actually, the only change you need to make is

```
    case 'SORT_CHANGE': {
      const { sortColumn, sortDirection } = action;

      return {
        ...state,
        sortColumn,
        sortDirection,
        currentPage: 1,
      };
    }
```

I will open a separate PR to fix the onChangePage in DataTable

jbetancur:
@JeremyEllingham I've merged in https://github.com/jbetancur/react-data-table-component/pull/168 with fixes/refactors to all the `onX` callbacks to use useEffects. This prevents having to tightly couple callbacks such as `onChangePage` to the actual sort function

If you'd like just go ahead and amend your PR with the state changes in my comments and I will merge in and cut a 3.0.2

PR #172, opened 2019-07-30
Reset pagination on table sort.

JeremyEllingham:
You're right that this is a better solution.  I didn't go straight for changing the reducer because I was running out of time at the end of the day and I was also half expecting that you would want the feature to be a prop-based conditional which would wrap the dispatch action.  Glad it isn't though.

Definitely get something into the Readme about specifically needing Node v10.15.0 though, that would have helped me a lot.


JeremyEllingham:
I had to overwrite my master branch to separate out the Search Box PR and that seems to have thrown code coverage for a loop, this PR doesn't have that scary red impact.

PR #173, opened 2019-07-30
Search box

JeremyEllingham:
Added an optional search box (set with boolean prop) to filter table data by standard string input.

Comes with documentation, a new Story, and the intelligence to only check against values actually displayed in the table rather than any that match the data set.  Doesn't have tests (although I can add them later if required) and I'm expecting some need for an extra useMemo in feedback because that seems to be a performance optimization (?) used on all the other data in this component but I haven't read up on that hook yet.  It might be able to use a bit of CSS as well to look a bit nicer.

### DIFF
--- a/src/DataTable/tableReducer.js
+++ b/src/DataTable/tableReducer.js
@@ -22,6 +22,7 @@ export function tableReducer(state, action) {
         sortColumn,
         selectedColumn,
         sortDirection,
+        currentPage: 1,
       };
     }
 
--- a/src/DataTable/tableReducer.js
+++ b/src/DataTable/tableReducer.js
@@ -47,6 +47,13 @@ export function tableReducer(state, action) {
       };
     }
 
+    case 'UPDATE_SEARCH_TEXT': {
+      return {
+        ...state,
+        searchText: action.text,
+      };
+    }
+
     case 'CHANGE_PAGE': {
       return {
         ...state,
--- a/README.md
+++ b/README.md
@@ -151,6 +151,11 @@ When the breakpoint is reached the column will be hidden. These are the built-in
 | onSort | func | no |  | callback to access the sort state when a column is clicked. returns ([column](https://github.com/jbetancur/react-data-table-component#columns), sortDirection, event) |
 | sortFunction | func | no |  | pass in your own custom sort function e.g. `(rows, field, direction) => {...yourSortLogicHere}. you must return an array |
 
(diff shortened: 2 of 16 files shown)


src/DataTable/tableReducer.js
```diff
@@ -22,6 +22,7 @@ export function tableReducer(state, action) {
         sortColumn,
         selectedColumn,
         sortDirection,
+        currentPage: 1,
```
jbetancur: Can open this as a separate PR as it is a different "feature"

JeremyEllingham: I did, here: https://github.com/jbetancur/react-data-table-component/pull/172

I presumed it would be merged by the time this one was, so I didn't bother extracting it from this one.

jbetancur: This is a bigger change and I want to get your sorting fix into a 3.0.3 release :)

I'll need some time to review this PR today or tomorrow

JeremyEllingham: No problem, the sorting fix was really what I needed.  The search box feature I already have production ready on my end implemented on top of the library, but since I'd already done it and it's an obvious feature I'm sure putting it in the main library will be useful for someone.

jbetancur: I thought about this a bit and instead of changing the source, I prefer the approach you took to add a custom search in the subHeader.  There are just so many ways someone would want to customize this that I think the subHeader `hook`  is the place to go. Additionally, we would need to handle the text input where people would like to use a 3rd party UI control; which only adds complexity to this library.

**Sidenote** and related to this, I've also considered column filtered fields where one could have an input on each column, but this suffers from the complexity issue above. It might be interesting though to allow people to place whatever component they want similar to `subHeaderComponent` for that scenario. and a column.filter: true.

jbetancur:
I thought about this a bit and instead of changing the source, I prefer the approach you took to add a custom search in the subHeader.  There are just so many ways someone would want to customize this that I think the subHeader `hook`  is the place to go. Additionally, we would need to handle the text input where people would like to use a 3rd party UI control; which only adds complexity to this library.

**Sidenote** and related to this, I've also considered column filtered fields where one could have an input on each column, but this suffers from the complexity issue above. It might be interesting though to allow people to place whatever component they want similar to `subHeaderComponent` for that scenario. and a column.filter: true.

PR #188, opened 2019-08-03
Pre-disabled Row

harshzalavadiya:
This PR will allow user to have row disabled from selection if `disabled: true` is passed in row


harshzalavadiya:
having check boxes as dumb component is good idea i will refector it and update this PR

src/DataTable/tableReducer.js
```diff
@@ -4,12 +4,12 @@ export function tableReducer(state, action) {
   switch (action.type) {
     case 'SELECT_ALL': {
       const allChecked = !state.allSelected;
-
+      const selectedRows = allChecked ? action.rows.reduce((acc, row) => (row.disabled ? acc : [...acc, row]), []) : [];
```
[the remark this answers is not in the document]
harshzalavadiya: okay so I have removed reduce and replaced it with simple filter by disabled property

src/DataTable/Checkbox.js
```diff
@@ -35,6 +36,7 @@ const Checkbox = ({
       aria-label={name}
       checked={checked}
       onChange={() => null} // prevent uncontrolled checkbox warnings -  we don't need onChange
+      disabled={(name !== 'select-all-rows') && disabled}
```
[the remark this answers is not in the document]
harshzalavadiya: as you suggested there was no need of checking for `select-all-row` so that's removed
and now it will show appropriate cursor `pointer` or `not-allowed`

harshzalavadiya:
Hi @jbetancur 
I have made changes for custom prop,
happy to make changes if you have suggestions/improvements

src/DataTable/tableReducer.js
```diff
@@ -4,12 +4,13 @@ export function tableReducer(state, action) {
   switch (action.type) {
     case 'SELECT_ALL': {
       const allChecked = !state.allSelected;
+      const selectedRows = action.rows.filter(({ disabled = false }) => !disabled);
```
jbetancur: If a user was using a custom `selectableRowsDisabledField` then `disabled` would break here

src/DataTable/Checkbox.js
```diff
@@ -20,7 +21,7 @@ const Checkbox = ({
   };
 
   const TagName = component;
-  const baseStyle = TagName !== 'input' ? componentOptions.style : baseCheckboxStyle;
+  const baseStyle = TagName !== 'input' ? componentOptions.style : { ...baseCheckboxStyle, cursor: disabled ? 'not-allowed' : 'pointer' };
```
jbetancur: hmm, this will cause `Checkbox` to re-render since the style Object is being recreated on every render pass.

src/DataTable/DataTable.js
```diff
@@ -267,6 +268,7 @@ const DataTable = memo(({
                         row={row}
                         columns={columnsMemo}
                         selectableRows={selectableRows}
+                        selctableRowsDisabledField={selctableRowsDisabledField}
```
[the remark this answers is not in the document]
harshzalavadiya: Whoops, I bliendly copy pasted

src/DataTable/tableReducer.js
harshzalavadiya: changes done!

src/DataTable/Checkbox.js
harshzalavadiya: I don't know how to avoid that and don't much have idea on `styled-components` but because of we have dynamic component styling might be recomputed anyway

let me know if you have any solution.

stories/DataTable/constants/sampleMovieData.js
```diff
@@ -33,6 +33,7 @@ export default [
   {
     id: 3,
     title: 'The Shawshank Redemption',
+    disabled: true,
```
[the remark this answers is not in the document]
harshzalavadiya: okay I have created new storybook example and removed it from selection example

harshzalavadiya:
> In addition to my notes and the spelling errors, I'd really like to see some tests.

for testing I don't have much experiance but I could try that once we are done building functionality

src/DataTable/Checkbox.js
jbetancur: You would memoize - in this case with the `useMemo` hook to cache the calculation. However, on second thought just leave it as it is and I'll refactor in a separate PR.

jbetancur: Additionally, I need to check if there is a performance benefit to memoizing

src/DataTable/TableRow.js
```diff
@@ -56,75 +56,80 @@ const TableRowStyle = styled.div`
...
-        onClick={handleRowClick}
-        className="rdt_TableRow"
-      >
-        {selectableRows && (
-          <TableCellCheckbox
-            name={`select-row-${row[keyField]}`}
-            row={row}
-          />
-        )}
+const TableRow = memo(
```
jbetancur: Can you restore the original formatting here? Looks like you just changed ` selectableRowsDisabledField: PropTypes.string.isRequired,` so that should be the only diff.

stories/DataTable/constants/sampleChocolates.js
```diff
@@ -0,0 +1,148 @@
+// Author https://github.com/schloerke/cranvasOLD/blob/master/files/data/chocolates.csv
```
jbetancur: I know you put some work into these tasty chocolates, but you can just map some disabled props in the story book using the existing data sets.

```
import data from '../constants/desserts';

// then outside the storybook component
const customData = data.map(datum => ({ ...datum, disabled: false }));
// Then arbitratily just disable some
customData[1].disabled = true;
customData[3].disabled = true;
```

stories/DataTable/Basic/BasicSelect.stories.js
```diff
@@ -34,6 +34,7 @@ const BasicSelectable = () => (
     columns={columns}
     data={data}
     selectableRows
+    selectableRowsDisabledField="notSelectable"
```
jbetancur: no need for this

src/DataTable/TableColCheckbox.js
```diff
@@ -14,9 +14,18 @@ const TableColStyle = styled(CellBase)`
...
 const TableCol = () => {
-  const { dispatch, data, selectedRows, allSelected, selectableRowsComponent, selectableRowsComponentProps } = useTableContext();
+  const {
+    dispatch,
+    data,
+    selectedRows,
+    allSelected,
+    selectableRowsComponent,
+    selectableRowsComponentProps,
+    selectableRowsDisabledField,
```
jbetancur: This would never work as it was not added to the context in `DataTable`

src/DataTable/TableRow.js
jbetancur: Actually, see this comment https://github.com/jbetancur/react-data-table-component/pull/188/files#r312010155

We can just pass this down using context and avoid prop drilling.

stories/DataTable/Basic/BasicSelect.stories.js
harshzalavadiya: okay, so you meant `disabled` can be default prop value for `selectableRowsDisabledField` instead of null as of now?

harshzalavadiya:
This branch is quiet outdated I'll fork fresh copy and make changes

harshzalavadiya:
Closing in favor of https://github.com/jbetancur/react-data-table-component/pull/210

PR #189, opened 2019-08-05
Add a pixel unit to the select arrow's top value

chanhman:
This PR fixes a CSS bug where the pagination's select arrow is not aligning because the top value was missing a unit of measurement.

### DIFF
--- a/src/DataTable/Select.js
+++ b/src/DataTable/Select.js
@@ -36,7 +36,7 @@ const SelectWrapper = styled.div`
   color: ${props => props.theme.pagination.fontColor};
 
   svg {
-    top: 1;
+    top: 1px;
     right: 0;
     color: ${props => props.theme.pagination.fontColor};
     position: absolute;


chanhman:
Screen Shot 2019-08-05 at 3 34 37 PM

jbetancur:
You're right that this should be a value. side note: in production mode, it gets removed since it's invalid css.

I would argue that it should be `top: 0:` instead or remote `top` altogether. Otherwise having it at 1px looks off-center.

src/DataTable/Select.js
```diff
@@ -36,7 +36,7 @@ const SelectWrapper = styled.div`
   color: ${props => props.theme.pagination.fontColor};
 
   svg {
-    top: 1;
+    top: 1px;
```
jbetancur: This is off-center. I think `top: 0;` will suffice

chanhman:
I agree. I only kept the 1 with the assumption that it was intended.

jbetancur:
The `Select` control styles to be cleaned up and so I addressed this in #200.

PR #219, opened 2019-08-27
refactor row height | fix checkbox css |  feature: row dense

### LINKED ISSUE #217: New Feature - Dense Table Rows
I am thinking of adding a new prop `dense` that densely compacts table rows vertically. You can also do this via themes but the use case might be common enough to warrant a new prop

jbetancur:
* adds a new feature for setting a themeable `denseHeight` to rows
* removed padding that was being applied to Cells
* fix css styles on Checkbox
* add new dense story
* closes #217


doverradio:
Hello, how do you style the checkbox?  My project has css coming from elsewhere and it's making the checkboxes quite large.  How to control checkbox size?

doverradio:
See please my [stackoverflow question here](https://stackoverflow.com/questions/72409272/react-data-table-component-how-to-style-checkbox-large-checkboxes-to-small).

PR #228, opened 2019-08-29
add class to selected row

Heydariiii128:
# add class to selected row
* for custom css style


Heydariiii128:
# add selected row style
# implement custom checkbox style:
preview: 
https://cdn1.imggmi.com/uploads/2019/8/29/949c821bd3180bc5ec3c34f33fb6beb6-full.png

* customCheckbox.scss:

.custom-checkbox {
	opacity: 0;
	width: 1px;
	height: 1px;
	visibility: hidden;
	pointer-events: none;
	position: absolute;

	+ label {
		display: inline-block;
		position: relative;
		width: 22px;
		height: 22px;
		background: #fcfcfc;
		border: 1px solid #eeeeee;
		@include border-radius(3px);
		margin: 0;
		cursor: pointer;

		&:before,
		&:after {
			content: '';
			position: absolute;
			display: inline-block;
			left: 0;
			top: 0;
			width: 100%;
			height: 100%;
		}
		&:before {
			background-color: transparent;
			opacity: 0;
			visibility: hidden;
			z-index: 0;
			@include border-radius(3px);
			@include transition(0.35s all ease);
		}
		&:after {
			z-index: 1;
			background-image: url('/check.svg');
			background-position: center center;
			background-repeat: no-repeat;
			background-size: 14px 11px;
			opacity: 0;
			visibility: hidden;
			@include transform(scale(0));
		}
	}
	&:checked + label {
		&:before,
		&:after {
			opacity: 1;
			visibility: visible;
			@include transition(0.4s all ease);
		}
		&:before {
			background-color: #ff9900;
		}

		&:after {
			@include transform(scale(1));
		}
	}
}

Heydariiii128:
thanks to all

jbetancur:
I appreciate the PR, but if you just want to add a custom checkbox you can just use the `selectableRowsComponent` together with `selectableRowsComponentProps`.

`selectableRowsComponent` allows you to pass whatever type of React component you want as long as it is based off a checkbox input

Readme:
https://github.com/jbetancur/react-data-table-component#row-selection

Also, the following storybook has an example:
https://jbetancur.github.io/react-data-table-component/?path=/story/ui-libraries--react-md

Otherwise, please write up some more details as to what your goals are use case etc.

**Side note:** not sure if your linting is not setup in your dev environment which is causing the build to fail. You may want to get that fixed or you can run `yarn lint` for a quick validation. Also,  when linting is off like this it's REALLY hard to tell what was changed in the code.

Heydariiii128:
that's right, I had better create a custom theme Instead of changing the original source. thanks for your full description.
I hope I can help your team better develop this project

sincerely...

PR #243, opened 2019-09-16
Fix example of memoized columns function

lenaggar:


jbetancur:
Good catch. Thank you!

PR #262, opened 2019-10-02
ExportTable Props

barisates:
exportTable - Show export button.
exportHeader - Add header to export file.
exportFileName - Default export file name.


jbetancur:
@barisates I appreciate the work you put into this PR! However, you can do this with the `subHeaderComponent` property or the `actions` property. 

You could instead adapt your work into a new storybook example showing how you would implement a table export using `subHeaderComponent`

PR #322, opened 2019-11-20
TypeScript Definitions

janhartmann:
Thanks for this great library!

I just started using it and figured out that there was not typings for TypeScript, so I went ahead made them for you. 😀

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,173 @@
+  import * as React from "react";
+  import { CSSProperties } from "styled-components";
+
+  export interface IDataTableProps<T> {
+    title?: React.ReactNode;
+    columns: IDataTableColumn<T>[];
+    data: T[];
+    keyField?: string;
+    striped?: boolean;
+    highlightOnHover?: boolean;
+    pointerOnHover?: boolean;
+    className?: string;
+    style?: CSSProperties;
+    responsive?: boolean;
+    customTheme?: IDataTableTheme;
+    disabled?: boolean;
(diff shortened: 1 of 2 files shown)


vladimirvorobiev:
nice !

index.d.ts
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  export interface IDataTableConditionalCellStyles<T> {
+    when: (row: T) => boolean;
+    style: CSSProperties;
+  }
+
+  export interface IDataTablePaginationOptions {
+    rowsPerPageText?: string;
+    rangeSeparatorText?: string;
```
jbetancur: missing `noRowsPerPage` boolean

jbetancur:
LOVE IT and thank you! Just one comment to address in IDataTablePaginationOptions

package.json
```diff
@@ -6,7 +6,8 @@
   "module": "dist/index.es.js",
   "browser": "dist/index.cjs.js",
   "files": [
-    "dist"
+    "dist",
+    "index.d.ts"
```
jbetancur: I'm still getting familiar with ts. Should this just go in dist? Do we need to add a "types" section in package.json?

janhartmann: I've added in `typings`, thanks for the highlight.

index.d.ts
janhartmann: Fixed and added more missing properties.

jbetancur:
Thank you! I'm going to cut this into a 4.0 release later today.

PR #323, opened 2019-11-21
remove flow | prefer typescript

jbetancur:


jbetancur:
Eventually moving to typescript I think so removing flow internals (we were not using flow anyway)

Issue #331, opened 2019-11-25
Does not recalculate the pagination position when the last item is deleted when paginationOnServer


sarathdr:
## Describe the bug
When the server-side pagination is active it does not calculate the pagination position when last item is deleted. The current page number is not decreasing. 

## To Reproduce
1. Go to [(code sandbox link)](https://codesandbox.io/s/beautiful-pasteur-16n73?fontsize=14&hidenavigation=1&theme=dark)
2. Go to second page
3. Click delete button
4. Check range below it shows 11 - 20 of 10, it should be 1-10 of 10

## Code Sandbox
[(code sandbox link)](https://codesandbox.io/s/beautiful-pasteur-16n73?fontsize=14&hidenavigation=1&theme=dark)

## Versions (please complete the following information)
 - React: 16.8.4
 - Styled Components: 4.4.1
 - Version: 4.0.0

## It seems the position should be recalculated when the total changes

```
  if(paginationServer) {
    React.useEffect(function () {
      if(paginationTotalRows > 0 && paginationServer) {
        var updatedPage = getNumberOfPages(paginationTotalRows, rowsPerPage);
        var recalculatedPage = recalculatePage(currentPage, updatedPage);
        handleChangePage(recalculatedPage);
      }
    }, [paginationTotalRows]);
  }
```

jbetancur:
Fixed by #337

PR #333, opened 2019-11-26
Pagination issue when pagination on server 

sarathdr:
[Issue](https://github.com/jbetancur/react-data-table-component/issues/331)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,3 +68,4 @@ storybook-static
 package-lock.json
 .DS_Store
 .vscode
+.idea
--- a/src/DataTable/DataTable.js
+++ b/src/DataTable/DataTable.js
@@ -125,7 +125,7 @@ const DataTable = memo(({
   const expandableRowsComponentMemo = useMemo(() => expandableRowsComponent, [expandableRowsComponent]);
   const handleRowClicked = useCallback((row, e) => onRowClicked(row, e), [onRowClicked]);
   const handleRowDoubleClicked = useCallback((row, e) => onRowDoubleClicked(row, e), [onRowDoubleClicked]);
-  const handleChangePage = page => dispatch({ type: 'CHANGE_PAGE', page, paginationServer });
+  const handleChangePage = useCallback(page => dispatch({ type: 'CHANGE_PAGE', page, paginationServer }));
 
   const sortedData = useMemo(() => {
     // server-side sorting bypasses internal sorting
@@ -195,6 +195,18 @@ const DataTable = memo(({
     const recalculatedPage = recalculatePage(currentPage, updatedPage);
 
     handleChangePage(recalculatedPage);
+  }
+
+  if (paginationServer) {
+    useEffect(() => {
+      if (paginationTotalRows > 0) {
+        const updatedPage = getNumberOfPages(paginationTotalRows, rowsPerPage);
+        const recalculatedPage = recalculatePage(currentPage, updatedPage);
+        if (currentPage !== recalculatedPage) {
+          handleChangePage(recalculatedPage);
+        }
+      }
+    }, [paginationTotalRows, currentPage, rowsPerPage, handleChangePage]);
   }
 
   const handleChangeRowsPerPage = newRowsPerPage => {
--- a/src/DataTable/DataTable.js
+++ b/src/DataTable/DataTable.js
@@ -125,7 +125,7 @@ const DataTable = memo(({
   const expandableRowsComponentMemo = useMemo(() => expandableRowsComponent, [expandableRowsComponent]);
   const handleRowClicked = useCallback((row, e) => onRowClicked(row, e), [onRowClicked]);
   const handleRowDoubleClicked = useCallback((row, e) => onRowDoubleClicked(row, e), [onRowDoubleClicked]);
-  const handleChangePage = useCallback(page => dispatch({ type: 'CHANGE_PAGE', page, paginationServer }));
+  const handleChangePage = page => dispatch({ type: 'CHANGE_PAGE', page, paginationServer });
 
   const sortedData = useMemo(() => {
     // server-side sorting bypasses internal sorting
--- a/src/DataTable/__tests__/DataTable.test.js
+++ b/src/DataTable/__tests__/DataTable.test.js
@@ -1407,6 +1407,45 @@ describe('DataTable::Pagination', () => {
     expect(onChangePageMock).toBeCalledWith(1, 1);
   });
 
+  test('should recalculate pagination position when the total number of items changes on pagination is set it on server', () => {
+    const mock = dataMock();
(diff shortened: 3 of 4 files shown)


jbetancur:
Thanks for taking a stab at fixing this.

I don't think you even need to re-calculate. Just detect if the paginationTotalRows changed then check the currentPage is > 1 and there are 0 records in the data.length or calculatedRows.length. then just move back a page

  if (pagination && paginationServer) {
    useDidUpdateEffect(() => {
      if (calculatedRows.length === 0 && currentPage > 1) {
        handleChangePage(currentPage - 1);
      }
    }, [paginationTotalRows]);
  }

I would also use `useDidUpdateEffect` since it doesn't fire on initial render

jbetancur:
```
  if (pagination && paginationServer && ) {
    useDidUpdateEffect(() => {
      if (calculatedRows.length === 0 && currentPage > 1) {
        handleChangePage(currentPage - 1);
      }
    }, [paginationTotalRows]);
  }
```

sarathdr:
> ```
>   if (pagination && paginationServer && ) {
>     useDidUpdateEffect(() => {
>       if (calculatedRows.length === 0 && currentPage > 1) {
>         handleChangePage(currentPage - 1);
>       }
>     }, [paginationTotalRows]);
>   }
> ```

`calculatedRows.length` this will be `> 0` since server should fetch the data for the last page? 
only the total changes?

src/DataTable/__tests__/DataTable.test.js
```diff
@@ -1407,6 +1407,45 @@ describe('DataTable::Pagination', () => {
...
+        onChangePage={onChangePageMock}
+        paginationPerPage={1}
+        paginationTotalRows={1}
+        paginationRowsPerPageOptions={[1, 2]}
+        onChangeRowsPerPage={onChangeRowsPerPageMock}
+        paginationServer
+        pagination
+      />,
+    );
+
```
sarathdr: @jbetancur Could you check the above props for re-render?

.gitignore
```diff
@@ -68,3 +68,4 @@ storybook-static
 package-lock.json
```
sarathdr: should we also commit `package-lock.json` ? It seems snapshot is failing because of that

jbetancur: I would stick with yarn and not include this file.

jbetancur:
caclulatedRows.length when using `paginationServer` should always be the same as data.length. It would probably be less confusing to just use data.length.

Since you are 'server-side" paging the calculatedRows.length will only be whatever the total length of the data loaded from you api is. So, if your paging for 10 records per page the length is 10.

What needs to be done is when the data.length reaches 0 then we trigger the actualy pagination back one page. e.g. you delete the last record on a page.

This is why `paginationTotalRows` matters when using paginationServer and why it should be set to whatever your pagination api returns as the total rows so that RDT can calculate what the total rows actually are instead of from data.length.

All of this works differently in client-side pagination since we have the entire dataset loaded which i why the code uses calculatedRows.length for those calculations (it uses Array.slice to give you the effect of pagination)

jbetancur:
```
 if (pagination && paginationServer) {
    useDidUpdateEffect(() => {
      if (data.length === 0 && currentPage > 1) {
        handleChangePage(currentPage - 1);
      }
    }, [paginationTotalRows]);
  }
```

.gitignore
sarathdr: Ah sorry I thought it uses npm

sarathdr:
> ```
>  if (pagination && paginationServer) {
>     useDidUpdateEffect(() => {
>       if (data.length === 0 && currentPage > 1) {
>         handleChangePage(currentPage - 1);
>       }
>     }, [paginationTotalRows]);
>   }
> ```

ok I will test again and resubmit

src/DataTable/DataTable.js
```diff
@@ -197,6 +197,18 @@ const DataTable = memo(({
     handleChangePage(recalculatedPage);
   }
 
+  if (paginationServer) {
```
janhartmann: Please see https://reactjs.org/docs/hooks-overview.html#rules-of-hooks

> Only call Hooks at the top level. Don’t call Hooks inside loops, conditions, or nested functions.

jbetancur: @janhartmann - I forgot about that and totally agree. I think we need to re-think the design the implementation of useEffect hooks in DataTable. I also think we need to move these hooks into their own files/logic.

jbetancur:
https://github.com/jbetancur/react-data-table-component/issues/335

Empty right now, but a placeholder for refactor work

src/DataTable/DataTable.js
sarathdr: I thought about it as well, @jbetancur need to clean up that

jbetancur:
Strange thing is the hooks linter does not catch it

jbetancur:
I just mocked up an actual delete story and it seems to be working. @sarathdr  your logic solves the issue. Just need to get that test working. I'll have a look 

```

  useDidUpdateEffect(() => {
    if (pagination && paginationServer && data.length === 0 && currentPage > 1) {
      handleChangePage(currentPage - 1);
    }
  }, [paginationTotalRows]);
```

jbetancur:
Actually, if you want to just remove the tests/snapshot updates I can merge this in. I can add the test in the refactor I am doing

janhartmann:
Remember to put the “pagination” in the dependency array. Good rule is that all used variables within the hook needs to go into the dependency array, so if any of these change it will re-compute.

janhartmann:
Or not sure if needed in this case

sarathdr:
> Actually, if you want to just remove the tests/snapshot updates I can merge this in. I can add the test in the refactor I am doing

OK I will push after that

jbetancur:
I created a codesandbox of what I am testing locally. It will server as a validation once 4.0.1 is cut

https://codesandbox.io/s/react-data-table-server-side-pagination-delete-xmdju

1. Nav to last page
2. delete both records
3. should nav to page 1

sarathdr:
> I created a codesandbox of what I am testing locally. It will server as a validation once 4.0.1 is cut
> 
> https://codesandbox.io/s/react-data-table-server-side-pagination-delete-xmdju
> 
> 1. Nav to last page
> 2. delete both records
> 3. should nav to page 1

jbetancur:
@janhartmann not needed in this case. While it is good practice in general, this sort of update is specific to paginationTotalRows changing (ie, a remote delete where the API returns some different row count. eg we don't want the page check to fire if say pagination is togged to false

jbetancur:
@sarathdr Make sure you rebase form master as I just made some ordering changes to useEffect

sarathdr:
> @sarathdr Make sure you rebase from master as I just made some ordering changes to useEffect

The pull request is closed, has this been merged ?

PR #337, opened 2019-11-28
Pagination issue when pagination on server

sarathdr:
[Issue](https://github.com/jbetancur/react-data-table-component/issues/331)


jbetancur:
I pulled this PR down and it looks like the pagination nav fires every time there is a change - which we don't want. Also, I don't think you have to do the recalculation instead, you can just determine if there are 0 records on the current page then go back one as long as the page is greater than 1. Also, we need to make sure this only activates when pagination/paginationServer.

The following seems to work based on the previous PR discussion with some modifications after testing.

```js
  useDidUpdateEffect(() => {
    if (pagination && paginationServer && data.length === 0 && currentPage > 1) {
      handleChangePage(currentPage - 1);
    }
  }, [currentPage, data.length, handleChangePage, pagination, paginationServer, paginationTotalRows]);
```

sarathdr:
> I pulled this PR down and it looks like the pagination nav fires every time there is a change - which we don't want. Also, I don't think you have to do the recalculation instead, you can just determine if there are 0 records on the current page then go back one as long as the page is greater than 1. Also, we need to make sure this only activates when pagination/paginationServer.
> 
> The following seems to work based on the previous PR discussion with some modifications after testing.
> 
> ```js
>   useDidUpdateEffect(() => {
>     if (pagination && paginationServer && data.length === 0 && currentPage > 1) {
>       handleChangePage(currentPage - 1);
>     }
>   }, [currentPage, data.length, handleChangePage, pagination, paginationServer, paginationTotalRows]);
> ```
> 
> Note here we should use `useDidUpdateEffect` over `useEffect,` which is just a custom hook that makes sure that `useEffect` only fires after the first render.

I am not sure the above solution works, for server-side pagination, it depends on the total number of records, so you have two options when you delete an item 

Option 1: Send a request to delete an item and update the state on client-side ( - remove the item from the store and decrease the total)

Your solution works with Option 1. But there are issues

1. It can affect the row order,  for example, if you remove an item from the middle of the records, we might need to pull one item from next page, which is not possible as we dont have next page records. There might be other potential issues with client state update.

2. Option 2 to reload the current page records from the server. Which is more consistent, but your solution won't work because it still depends on `data.length`. Server loads data and total for the current page then data.length will not be equal to `0`

I will replace `useEffect` with `useDidUpdateEffect` , reacalculation is required as we use pageTotal here, I think is not an expensive operation. 

I am happy to discuss it over skype or something if it requires more clarity

jbetancur:
The sandbox will fail because it does not point to our changes. It was more for copying the logic to a local dev environment as I am doing with my testing.

It might be easier if we just step through to make sure I'm not missing something 😉 

## Option 2
* Let's say we have 2 pages with a total of 12 records and we navigate to page 2 
GET https://reqres.in/api/users?page=2&per_page=10 (returns a total count of 12, plus our data)
Page 2 now has a `data.length` of  2 and `paginationTotalRows` is 12
* I delete one record which requires 2 operations as per Option 2
DELETE https://reqres.in/api/users/12
GET https://reqres.in/api/users?page=2&per_page=10 (to "refresh" the data state)
Page 2 now has a `data.length` of  1 and `paginationTotalRows` from the GET should now be 11 
* I delete the last record on page 2 which again requires 2 operations as per Option 2
DELETE https://reqres.in/api/users/11
GET https://reqres.in/api/users?page=2&per_page=10 (to "refresh" the data state)
Page 2 now has a `data.length` of  0 and `paginationTotalRows` from the GET should now be 10 
page

Our effect to move back ONE page will now run because either props data.length or paginationTotalRows have changed. 

Since 0 records are returned and we are still on page 2 `handleChangePage(1)` is called which in turns calls: GET https://reqres.in/api/users?page=1&per_page=10 to page us back to page 1
```
  useDidUpdateEffect(() => {
   // 2. the api returns 0 records and we on a page greater than 1, so go back 1 page
    if (pagination && paginationServer && data.length === 0 && currentPage > 1) {
      handleChangePage(currentPage - 1);
    }
// 1. on the last delete we refresh the data which is now 0 so the effect triggers
  }, [currentPage, data.length, handleChangePage, pagination, paginationServer, paginationTotalRows]);
```

*keep in mind that our test REST API when you are back on page one will show 12 records because it doesn't actually delete the data but the count would otherwise be 10.*

sarathdr:
This will trigger an extra request to the server? ....even if the client knows that there is no records will be returned from the server( +1 db select) ?

jbetancur:
I think the API call would happen either way no? Your option 2 delete logic would be something like this:

```js
  const handleDelete = row => async () => {
    await axios.delete(`https://reqres.in/api/users/${row.id}`);
    const response = await axios.get(`https://reqres.in/api/users?page=${currentPage}&per_page=${perPage}`);

    setData(response.data.data);
    setTotalRows(response.data.total);
  };

```
I think the extra API call is fine because you are asking the server to return you the actual total rows and data - which even if you deleted the record could have changed if a user elsewhere added a record while you were deleting. In other words, it keeps the table synced with your server state even if it took an API call that returned 0 records.

`onChangePage` simply fires whatever logic you want whether the page is changed by clicking the page button or if data.length === 0 && currentPage > 1

sarathdr:
No we need only two calls, in your case its three calls 

1. to delete
2. loads last page 
3. loads again the second last page

If we check local page count and it is the last item, on client-side we request the current page, 

Datatable sync will not work anyway in the current implementation, for example, if you can 11 records from another tab, the current tab will not go to next page (we need to test this usecase as well, I am not sure it will show the extra pages when the total is changed as the pages will not be recalculated)

jbetancur:
What is bugging me is that"
```
      if (currentPage !== recalculatedPage) {
        handleChangePage(recalculatedPage);
      }
```

fires multiple times, that is the check is performed constantly when you console it out. We only care about `paginationTotalRows` changing so let's just only include that updateProp and also add checks that pagination and paginationServer are true:


```js
  useDidUpdateEffect(() => {
    if (pagination && paginationServer && paginationTotalRows > 0) {
      const updatedPage = getNumberOfPages(paginationTotalRows, rowsPerPage);
      const recalculatedPage = recalculatePage(currentPage, updatedPage);

      if (currentPage !== recalculatedPage) {
        handleChangePage(recalculatedPage);
      }
    }
  }, [paginationTotalRows]);
```

Otherwise, if you have other use cases I am missing please let's get some code and show them, but I would like to get a release out asap.

janhartmann:
Would like to have this out as well, hitting the same issue now and had to revert the version update.

sarathdr:
Ok I will update and push it soon

src/DataTable/__tests__/DataTable.test.js
```diff
@@ -1407,6 +1407,45 @@ describe('DataTable::Pagination', () => {
...
+    const onChangeRowsPerPageMock = jest.fn();
+    const { container, rerender } = render(
+      <DataTable
+        data={mock.data}
+        columns={mock.columns}
+        onChangePage={onChangePageMock}
+        paginationPerPage={1}
+        paginationTotalRows={2}
+        paginationRowsPerPageOptions={[1, 2]}
+        onChangeRowsPerPage={onChangeRowsPerPageMock}
```
jbetancur: onChangeRowsPerPage is not doing anything? I'll remove this in another PR

jbetancur:
Closes #333

src/DataTable/__tests__/DataTable.test.js
```diff
@@ -1407,6 +1407,45 @@ describe('DataTable::Pagination', () => {
     expect(onChangePageMock).toBeCalledWith(1, 1);
   });
 
+  test('should recalculate pagination position when the total number of items changes on pagination is set it on server', () => {
```
jbetancur: will fix this sentance

jbetancur:
4.0.2 has been released

PR #343, opened 2019-12-02
Unique IDs and ARIA attributes

AtomicNoggin:
- Add unique id to all elements so multiple data tables can exists on the same page
 - Added better aria attribute support to help with a11y -- including  screen reader only labels for selector and expander headers.
 - Moved ExpanderRow into parent TableRow so it will semantically be considered a cell by screen readers.
- Updated tests to handle new id strategy.

### DIFF
--- a/src/DataTable/ExpanderRow.js
+++ b/src/DataTable/ExpanderRow.js
@@ -1,6 +1,6 @@
-import React, { Children, cloneElement } from "react";
-import PropTypes from "prop-types";
-import styled from "styled-components";
+import React, { Children, cloneElement } from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
 
 // Make "data" available on our any child component
 // eslint-disable-next-line arrow-body-style
@@ -23,16 +23,18 @@ const ExpanderRow = ({ data, children, id, index }) => (
 );
 
 ExpanderRow.propTypes = {
+  id: PropTypes.any.isRequired,
+  index: PropTypes.number.isRequired,
   data: PropTypes.object,
   children: PropTypes.oneOfType([
     PropTypes.arrayOf(PropTypes.node),
-    PropTypes.node
-  ])
+    PropTypes.node,
+  ]),
 };
 
 ExpanderRow.defaultProps = {
   data: {},
-  children: null
+  children: null,
 };
 
 export default ExpanderRow;
--- a/src/DataTable/ExpanderRow.js
+++ b/src/DataTable/ExpanderRow.js
@@ -1,6 +1,6 @@
-import React, { Children, cloneElement } from 'react';
-import PropTypes from 'prop-types';
-import styled from 'styled-components';
+import React, { Children, cloneElement } from "react";
+import PropTypes from "prop-types";
+import styled from "styled-components";
 
 // Make "data" available on our any child component
 // eslint-disable-next-line arrow-body-style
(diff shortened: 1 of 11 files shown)


jbetancur:
Thanks for addressing the accessibility of RDT (or lack thereof) with this PR.

This is an insanely busy week so I haven't had time to check out this PR and test locally. I left a couple of comments on things that stand out.

src/DataTable/ExpanderRow.js
```diff
@@ -9,22 +9,22 @@ const renderChildren = (children, data) => {
...
 
 const ExpanderRowStyle = styled.div`
+  flex: 0 0 100%;
   width: 100%;
   box-sizing: border-box;
   color: ${props => props.theme.expander.fontColor};
   background-color: ${props => props.theme.expander.backgroundColor};
 `;
 
-const ExpanderRow = ({
```
jbetancur: We should preserve the original formatting to keep not only to keep the diff clean but to be consistent with other components

src/DataTable/TableRow.js
```diff
@@ -50,12 +52,15 @@ const pointerCSS = css`
...
 const TableRowStyle = styled.div`
   display: flex;
+  flex-wrap: wrap;
   align-items: stretch;
   align-content: stretch;
   width: 100%;
   box-sizing: border-box;
-  min-height: ${props => (props.dense ? props.theme.rows.denseHeight : props.theme.rows.height)};
-  ${props => (props.theme.rows.spacing === 'spaced' ? spacedRowsCSS : defaultRowsCSS)};
+  min-height: ${props =>
```
jbetancur: We should preserve the original formatting to keep not only to keep the diff clean but to be consistent with other components

jbetancur: Also, we don't want to change this since this is not relevant to the PR

src/DataTable/TableRow.js
```diff
@@ -64,60 +69,92 @@ const TableRowStyle = styled.div`
...
-      }
-    }
-  }, [disableRowClick, expandOnRowDoubleClicked, expandableRows, handleExpanded, onRowDoubleClicked, row]);
-
-  const extendedRowStyle = getConditionalStyle(row, conditionalRowStyles);
-
-  return (
-    <>
+const TableRow = memo(
+  ({
```
jbetancur: We should preserve the original formatting to keep not only to keep the diff clean but to be consistent with other components

src/DataTable/TableRow.js
```diff
@@ -64,60 +69,92 @@ const TableRowStyle = styled.div`
...
+        // use event delegation allow events to propagate only when the element with data-tag ___react-data-table-allow-propagation___ is present
+        if (e.target && e.target.getAttribute('data-tag') === STOP_PROP_TAG) {
+          onRowClicked(row, e);
+
+          if (!disableRowClick && expandableRows && expandOnRowClicked) {
+            handleExpanded();
+          }
+        }
+      },
+      [
```
jbetancur: Again keep the original formatting

src/DataTable/TableRow.js
```diff
@@ -64,60 +69,92 @@ const TableRowStyle = styled.div`
...
+      e => {
+        if (e.target && e.target.getAttribute('data-tag') === STOP_PROP_TAG) {
+          onRowDoubleClicked(row, e);
+          if (!disableRowClick && expandableRows && expandOnRowDoubleClicked) {
+            handleExpanded();
+          }
+        }
+      },
+      [
+        disableRowClick,
```
jbetancur: keep the original formatting

src/DataTable/ExpanderRow.js
```diff
@@ -9,22 +9,22 @@ const renderChildren = (children, data) => {
 };
 
 const ExpanderRowStyle = styled.div`
+  flex: 0 0 100%;
```
jbetancur: Is this relevant to accessibility or is this a fix to expander?

src/DataTable/ExpanderRow.js
```diff
@@ -1,6 +1,6 @@
-import React, { Children, cloneElement } from 'react';
-import PropTypes from 'prop-types';
-import styled from 'styled-components';
+import React, { Children, cloneElement } from "react";
```
jbetancur: single quotes

PR #365, opened 2019-12-12
Made `onRowExpandToggled` optional

janhartmann:
This prop is not required, fixed typings.


jbetancur:
Thanks for catching that @janhartmann !

PR #368, opened 2019-12-13
fix(onRowExpandToggled): Make onRowExpandToggled prop optional

virtuoushub:
See: 8cb74a878f1a2f9cc9be4f34ee672c3d935a6dda

Not sure if it was intentional or not, but the `onRowExpandToggled` prop is currently not optional.

My IDE's TypeScript tooling is yelling at me to include it where I do not think I should have to.

This commit makes it's inclusion as a property optional.


virtuoushub:
whoops, #365 was already made,

this duplicates that. closing this pr.

Issue #363, opened 2019-12-09
[6.0.0] - remove memoize-one and let people import their own


jbetancur:
currently, the wonderful `memoize-one` is bundled with RDT and exportable - instead this should be opt in.

* remove  memoize-one
* remove the export
* update the readme to point to  memoize-one

Issue #357, opened 2019-12-08
[6.0.0] - Grid lines


jbetancur:
Depends on #355 
Grid lines - either through theming or via a helper prop

blacksmoke26:
Yeah, That will be cool. I am really impressed by this library. :)

jbetancur:
Instead of a prop, this will actually be implemented via the new `customStyles` prop with accompanying storybook examples

blacksmoke26:
Yes, that's the better approach! Since it's related to the styles, adding it to `customStyles` is the best decision.

jbetancur:
Implemented in release/v6 branch

Issue #355, opened 2019-12-08
[6.0.0] - Theming Refactor to use style objects/css-in-js


jbetancur:
Refactor theming to use styled objects. Currently, theming is very confusing so why not let you use css-in-jss style formatting for theming RDT?! Note that this will be a breaking change as theme will need to be re-thought from the ground up

PR #380, opened 2019-12-18
Release/v6

jbetancur:
## Breaking
* #355 `customStyles`, `theme`, dark mode and ability to create themes using `createTheme` | removal of `customTheme`
* #363 - no longer bundle in memoize-one library - let users chose their own.

## Features
* #357 add grid rows via `customStyles`
* #359 column selectors now accept a callback e.g. `selector: row => row.field` is equal to `selector: 'field'`


blacksmoke26:
Does any example contain the latest changes? A showcase?

jbetancur:
Look under the custom styling and the theming storybook

Issue #395, opened 2019-12-22
Some of properties are missing on index.d.ts


kokutaro:
## Describe the bug
While using the react-data-table-component with TypeScript, I noticed following issue.
1. Property `conditionalRowStyles` is missing in `interface IDataTableProps<T>` 
2.  Property `conditionalCellStyles` in `interface IDataTableColumn<T>` should be `array`.

I suggest to correct the `index.d.ts` as below.

``` Typescript
export interface IDataTableProps<T> {
  ...
  conditionalRowStyles?:IDataTableConditionalRowStyles<T>[];
}
...
export interface IDataTableColumn<T> {
  ...
  conditionalCellStyles?: IDataTableConditionalCellStyles<T>[];
}
...
export interface IDataTableConditionalRowStyles<T> {
  when: (row: T) => boolean;
  style: CSSProperties;
}
```

kokutaro:
I have created the pull request on #396

jbetancur:
Fixed in #396

jbetancur:
Slated for 6.0.5 release

jbetancur:
Now available in 6.0.4!

PR #396, opened 2019-12-22
#395 Issue

kokutaro:


index.d.ts
```diff
@@ -31,7 +31,7 @@ export interface IDataTableProps<T> {
     rows: T[],
     field: string,
     sortDirection: 'asc' | 'desc'
-  ) => T[];
+) => T[];
```
jbetancur: nit pick - just move this back to 2 spaces

kokutaro: Thank you for reviewing! I have just corrected!

jbetancur:
Nice! merged. I will include this is a 6.0.05 or 6.1.0

PR #421, opened 2020-01-09
Docs update: pass a component to column name

clmcdonald:
Updated the docs to make clear that you can pass a component to the column name field

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ Nothing new here - we are using an array of object literals and properties to de
 
 | Property | Type   | Required | Example                                                                                                       |
 |----------|--------|----------|---------------------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


jbetancur:
Thanks for the PR! So currently the interface for name only accepts a string but that's just if you are using typescript. Otherwise, with javascript, you can pass whatever you want since it is dynamically typed

```
export interface IDataTableColumn<T> {
  name: string;
...

```

I never thought of passing a component to a column name, though, but I like it. I'll go ahead and merge this is and I'll open a separate PR to add it to the types. It should probably also accept a number

clmcdonald:
Oh ok. Yeah my use case was I wanted to add a Tooltip to the column name so I just passed a component to see if it would work. I am not using TypeScript so that explains it

PR #443, opened 2020-01-29
Make all customStyles interface fields optional.

mute---:
Due to incorrect type definition it's unable to specify customStyles object as you encounter "fields not found" exception when using typescript.


jbetancur:
Good catch and thank you

PR #472, opened 2020-02-23
add last affected rows to state.

ahmed-abdelmotey:
This will enable users to handle actions depending on the selection of the rows.

### DIFF
--- a/src/DataTable/tableReducer.js
+++ b/src/DataTable/tableReducer.js
@@ -8,6 +8,7 @@ export function tableReducer(state, action) {
       return {
         ...state,
         allSelected: allChecked,
+        lastAffectedRows: action.rows,
         selectedCount: allChecked ? action.rows.length : 0,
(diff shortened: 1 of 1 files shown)


ahmed-abdelmotey:
Please letme know if there is some changes needed before I update the test files assertions if needed.

jbetancur:
Thanks for the PR! I’ve been a bit busy so have not had a chance to review. I’ll spend some time on it this week.

Off the top of my head would it be easier to just have a new callback that gives you the current row that was selected?

jbetancur:
@ahmed-abdelmotey can you provide an example of his this is used in practice?

ahmed-abdelmotey:
Thank you for your patience, 


the scenario is: 
if you have an array of data and a property is used to determine if the row is selected or no, once you update the selection in the table this should reflect on the state of the parent component, there is no way to find out the affected row unless you loop over all the items to determine which one was selected or un selected

PR #479, opened 2020-03-06
Apply row & column className to TableCell

jipema:
I needed specific class attribute for some cells.

Now each [column definition](https://github.com/jbetancur/react-data-table-component#columns) can have a `className` attribute that will be passed to the TableCell.

PS: indentation diffs because of Prettier + .editorconfig to match quotes style

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,6 +9,7 @@ root = true
 # Change these settings to your own preference
 indent_style = space
 indent_size = 2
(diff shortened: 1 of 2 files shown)


jbetancur:
Do you have an example of how you would use this so I can better understand?

Currently, you can use columns[].styles or conditionalStyles to override a cells styling.

Issue #481, opened 2020-03-11
pagination: expose the native Pagination.js file


justiceappiah-cult:
There is no way you can wrap your custom component over the native Pagination file found here https://github.com/jbetancur/react-data-table-component/blob/master/src/DataTable/Pagination.js

By exposing the native pagination via exports, you can:
- wrap a custom component over it
- reuse this same component in two different locations (eg. top and bottom of page)

## Describe alternatives you've considered
Without exposure, you'd have to use react refs to target the original paginator component, which is quite difficult and complex

PR #482, opened 2020-03-11
[R] expose pagination file

justiceappiah-cult:
Handles https://github.com/jbetancur/react-data-table-component/issues/481

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,6 +1,7 @@
 import DataTable from './DataTable/DataTable';
+import Pagination from './DataTable/Pagination';
 import { createTheme, defaultThemes } from './DataTable/themes';
 
-export { defaultThemes, createTheme };
+export { defaultThemes, createTheme, Pagination };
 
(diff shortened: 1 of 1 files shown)


jbetancur:
Thanks for the pr! Can you show an example of how You would wrap the native component/use case?

justiceappiah-cult:
@jbetancur In the scenario below, we want the Pagination to be duplicated and place at both the bottom and top our screen.

We can do it this way

```

import { Paginate as NativePaginate } from "react-data-table-component."

const SamplePage = () => 
   <DataTable
          pagination
          paginationComponent={OurShowCasePaginationComponent}
        />

const OurShowCasePaginationComponent = (props) => 
   <div>
       <TopOfOurScreen><NativePaginate {...props}/></TopOfOurScreen>
       <BottomOfOurScreen><NativePaginate {...props}/></BottomOfOurScreen>
  </div>


```

Issue #524, opened 2020-04-06
Multi-page Row Selection: Persist selected rows across pages when using serverPagination


MaxInertia:
## Feature Check list

- [x] Agree to the [Code of Conduct](https://github.com/jbetancur/react-data-table-component/blob/master/CODE-OF-CONDUCT.md)
- [x] Read the README to ensure the feature is not already present
- [x] You read [Creating Issues, Features and Pull Requests](https://github.com/jbetancur/react-data-table-component/issues/387)
- [x] Considered the value versus complexity for all users of the library as well as library maintenance
- [x] Considered if this can be a storybook or documentation example

## Is your feature request related to a problem? Please describe

Apparent inability to implement multi-page selection using this table.

Please let me know if I'm missing something obvious.

## Describe the solution you'd like

A new prop for the Data Table that prevents clearing of selected rows when the page or sort is changed **when using `serverPagination`.** Possibly a prop for each of them. This prop would be optional with the default being false so existing programs are not modified.

These two cases in the reducer would be changed in the following way:

Note the addition of `selectedRowsPersist`

```
case 'SORT_CHANGE': {
      const { sortColumn, sortDirection, selectedColumn, selectedRowsPersist, pagination, paginationServer } = action;

      return {
        ...state,
        sortColumn,
        selectedColumn,
        sortDirection,
        currentPage: 1,
        // when using server-side paging reset selected row counts when sorting
        ...pagination && paginationServer && !selectedRowsPersist && ({
          allSelected: false,
          selectedCount: 0,
          selectedRows: [],
        }),
      };
    }

case 'CHANGE_PAGE': {
      const { page, paginationServer, selectedRowsPersist } = action;
      return {
        ...state,
        currentPage: page,
        // when using server-side paging reset selected row counts
        ...paginationServer && !selectedRowsPersist ({
          allSelected: false,
          selectedCount: 0,
          selectedRows: [],
        }),
      };
    }
```

Alternative prop names:
- `persistSelectedRowsOnPageChange` & `persistSelectedRowsOnSortChange`

## Describe alternatives you've considered

- Supplying more information via the `onSelectedRowsChange` method that provides information required to implement multi-page selection in the user space. (**More work than proposed solution as far as I can tell**)

[...]

PR #525, opened 2020-04-07
Add option to persist selections on page&sort change w/paginationServer

MaxInertia:
Implements #524 

### New prop
```
paginationServerOptions: {
  persistSelectedOnSort: bool,
  persistSelectedOnPageChange: bool,
}
```
Defaults are `false`, this prevents behavior of existing programs from changing.

### Conflicts
Some changes are close to changes in #520 so merge conflicts will have to be resolved. If this PR is accepted I recommend merging #520 first, then I can resolve the conflicts again for a final approval & merge.


jbetancur:
Nice! I'll have a bit of a crazy schedule, but I'll take a look this week and run it locally.  I left a couple of comments

index.d.ts
```diff
@@ -77,6 +77,8 @@ export interface IDataTableProps<T> {
     selectedCount: number;
     selectedRows: T[];
   }) => void;
+  persistSelectedRowsOnSortChange?: boolean;
```
jbetancur, line 80: I really hate these long property names and having to manage them, (it's my fault 😉 ). When I created this lib I didn't envision it growing like this. 

I'm planning on a next version(v7 or v8) to move some of these to an options object so this would be a good start

paginationServerOptions: { persistSelectedOnSort: bool }

same probably for `persistSelectedRowsOnPageChange` below. 

We can set the defaults in the proptypes file


paginationServer
paginationServerOptions

jbetancur:
We should also add some tests to this if possible, but it may be hard to do so since it;'s hard to tell what is not selected without mock/spying on the chang handlers for onSelectedRowsChange etc

index.d.ts
MaxInertia: I totally agree. Modified to the following:
```
paginationServerOptions: {
  persistSelectedOnSort: bool,
  persistSelectedOnPageChange: bool,
}
```

`onPageChange` is more meaningful than `onPage`, so I went with that. If you're not fond of that though I can change it.

src/DataTable/__tests__/DataTable.test.js
```diff
@@ -1729,6 +1729,104 @@ describe('DataTable::Pagination', () => {
...
+        selectableRows
+        onSelectedRowsChange={onSelectedRowsChange}
+      />,
+    );
+
+    fireEvent.click(container.querySelector('input[name="select-all-rows"]'));
+    expect(onSelectedRowsChange).toBeCalledTimes(1);
+    fireEvent.click(container.querySelector('button#pagination-next-page'));
+    expect(onSelectedRowsChange).toBeCalledTimes(1);
+  });
```
MaxInertia, line 1775: Verifies the added prop `persistSelectedOnPageChange` can prevent clearing selections on page change

src/DataTable/__tests__/DataTable.test.js
```diff
@@ -1729,6 +1729,104 @@ describe('DataTable::Pagination', () => {
...
+      />,
+    );
+
+    fireEvent.click(container.querySelector('input[name="select-all-rows"]'));
+    expect(onSelectedRowsChange).toBeCalledTimes(1);
+
+    fireEvent.click(container.querySelector('div[id="column-some.name"]'));
+    expect(onSortMock).toBeCalled();
+    expect(onSelectedRowsChange).toBeCalledTimes(2);
+  });
```
MaxInertia, line 1801: Verifies default behavior remains the same on sort

src/DataTable/__tests__/DataTable.test.js
```diff
@@ -1729,6 +1729,104 @@ describe('DataTable::Pagination', () => {
...
+      />,
+    );
+
+    fireEvent.click(container.querySelector('input[name="select-all-rows"]'));
+    expect(onSelectedRowsChange).toBeCalledTimes(1);
+
+    fireEvent.click(container.querySelector('div[id="column-some.name"]'));
+    expect(onSortMock).toBeCalled();
+    expect(onSelectedRowsChange).toBeCalledTimes(1);
+  });
```
MaxInertia, line 1828: Verifies the added prop `persistSelectedOnSort` can prevent clearing selections on sort

jbetancur:
This is looking pretty good!  I did cause a few merge conflicts, especially in tableReducer.  You can do something like these one-liners to integrate your changes.

SORT_CHANGE
` const clearSelectedOnSort = (pagination && paginationServer && !persistSelectedOnSort) || sortServer || visibleOnly;`

and for CHANGE PAGE
`const clearSelectedOnPage = (paginationServer && !persistSelectedOnPageChange) || visibleOnly;`

src/DataTable/tableReducer.js
```diff
@@ -41,21 +41,20 @@ export function tableReducer(state, action) {
         ...state,
         selectedCount: selectedRows.length,
         allSelected: selectedRows.length === rows.length,
-        selectedRows,
+        selectedRows: [...state.selectedRows, ...selectedRows],
```
jbetancur, line 44: I missed this one. What was the reason for merging these?
 Versus just replacing?

jbetancur:
One more comment on tableReducer and we are good to go

src/DataTable/tableReducer.js
MaxInertia: Previously this `useEffect` replaced the selections with the subset of `data` that is filtered out by `selectableRowSelected`. `data` is only the current page when using `paginationServer`. So this change is made to avoid wiping out selections from other pages.

https://github.com/MaxInertia/react-data-table-component/blob/0714517d832bc805d0120d6a29c2d7f8f6fb9d98/src/DataTable/DataTable.js#L162-L170

I'm glad you brought this up though, there's a big problem with this: There are situations where we'll end up with duplicate rows in `selectedRows` with a naive merging. Only those that aren't already in `state.selectedRows` need to be merged.

MaxInertia: @jbetancur What do you think about something like this?

```
[...state.selectedRows, ...selectedRows.filter(row => !state.selectedRows.includes(row))]
```

My only concern is how this will fare when `state.selectedRows` and `selectedRows` are large...

jbetancur: The code above the worst case in O(n+m) - which is not horrible but a case that I would prefer to avoid. I know that even if someone has say ~1000 rows I am pretty sure the DOM will run into performance issues before this code (in which case you need to page), but still... it feels weird.

Aside from a for loop which would be negligible in terms of performance, there is Set.

```js
 Array.from(new Set(array))
```
The big gotcha with Set is it will need to be polyfilled for IE11 users so will need to figure that out.

https://stackoverflow.com/questions/39007637/javascript-set-vs-array-performance


Another solution could be to only perform the expensive operation when persistSelectedOnX where we can merge arrays but in any case we default to replace selectedRows to preserve the current performance.


Sorry for the long comment but this is making me think what is this use case looks like in your implementation? Typically UX wise it is recommended to only allow selections on visible items where this is more edge case of server-side pagination to make up for the limits of server-side pagination

MaxInertia: >Typically UX wise it is recommended to only allow selections on visible items where this is more edge case of server-side pagination to make up for the limits of server-side pagination

This isn't a necessary limitation with server-side pagination in general, it's a limitation with server-side pagination on this table. This change brings the behavior of the table when using server-side pagination in-line with it's behavior when not using server-side pagination.

Currently to get selection working on multiple pages we're forced to query for all our data upfront, which is wasteful.

>Another solution could be to only perform the expensive operation when persistSelectedOnX where we can merge arrays but in any case we default to replace selectedRows to preserve the current performance.

This might be the best option right now. Although, my code in the last comment is actually `O(nm)` since each of the previous selections results in an iteration over the new selections.

jbetancur: Yea O(nm) is correct, my error.. O(n+m) might have been better if it were possible.

You have a valid point on the limitation. Another option could be the cache the rows as off page rows and pass the merging off, but perhaps that adds more complexity at the cost of pre-optimization. If someone is paging why would they page 10000 rows (I’m sure someone will log that issue now :))

Anyway I just want to weigh all the options here.

Let’s do this. Just add the code to only to the expensive calculation when this option is specified and add the test as needed to check that the rows are cached on server page and let’s get this merged in.

Sidebar: It would be cool to have Some sample code for this for a working storybook example. Totally not needed for this pr though.

src/DataTable/tableReducer.js
```diff
@@ -35,27 +35,28 @@ export function tableReducer(state, action) {
     }
 
     case 'SELECT_MULTIPLE_ROWS': {
-      const { selectedRows, rows } = action;
+      const { selectedRows, rows, mergeSelections } = action;
+      const selections = !mergeSelections ? selectedRows :
```
jbetancur, line 39: [nit] ]This should just be 

```js
      const selections = mergeSelections 
             ? [...state.selectedRows, ...selectedRows.filter(row => !state.selectedRows.includes(row))];
             : selectedRows :
```

the negation may make this core harder to read.
        
```

**Comment by jbetancur** [comment]
I did some local testing with the following storybook: http://localhost:6006/?path=/story/pagination--server-side-using-hooks and row selection is not quite right.

When you select all the select all checkbox should be indeterminate. When you navigate to another page and click on a single row it should still be indeterminate. If all rows are selected then the select all should be checked.

1. We need to handle the indeterminate case. That is that if someone clicks on select all then we should not display that all rows are selected unless all the rows are selected.

2. We will need to handle the SELECT_SINGLE_ROW case as a single select also affects selected rows.

3. Need to fix the coverage/add a test

**Comment by jbetancur** [comment]
I did some local testing with the following storybook: http://localhost:6006/?path=/story/pagination--server-side-using-hooks and row selection is not right.

When you select all the select all checkbox should be indeterminate. When you navigate to another page and click on a single row it should still be indeterminate. If all rows are selected then the select all should be checked (but how can they be of our data is not all there?)

-  handle the indeterminate case. That is that if someone clicks on select all then we should not display that all rows are selected unless all the rows are selected.
    -  the SELECT_ALL_ROWS and SELECT_MULTIPLE_ROWS rowCount/ for example: `  const 
  rowCount = mergeSelections ? paginationTotalRows : data.length;`  This means that in 
  TableColCheckbox and TableCellCheckbox we need to also evaluate if the paginationServerOptions 
  are set. We can pass then down via context from DataTable

- how can we select all of the data rows that are not actually there? basically select all only selects all of the data loaded.  this kind of breaks this feature.

This is why I mentioned that there are limitations to server-side paging in this case. It's a lot of cognitive load (at least for me) to maintain.

I'll try to have a look later today to see if I can push up a branch with the points above

**Comment by jbetancur** [comment]
https://github.com/jbetancur/react-data-table-component/compare/MaxInertia-feature/524
and hit the compare button

There are some cases still not working:
- select all on a page > 1 breaks previously seletected `SELECT_ALL_ROWS` still suspect on how to handle this with our "row caching"
- select all only selects all items on the current page. unsure how to solve that given all the data is not loaded.

Perhaps the storybook http://localhost:6006/?path=/story/pagination--server-side-using-hooks is missing something that makes this work and we need another storybook for this specific case. These are sort of the integration tests...

**Comment by jbetancur** [comment]
Another thought, perhaps we should just ALWAYS disable select all when using these options - I don't see how to make it work

**Comment by MaxInertia** [comment]
> select all on a page > 1 break previously selected SELECT_ALL_ROWS still suspect on how to handle this with our "row caching"

Those weird cases should be resolved now.

... Could probably use more tests though, I have made `codecov` angry.

> select all only selects all items on the current page. unsure how to solve that given all the data is not loaded.

Yeah, nothing can be done about that. Select all in this case just means select all on the page, which is what it was before with `paginationServer` if I'm not mistaken?

> Another thought, perhaps we should just ALWAYS disable select all when using these options - I don't see how to make it work

Having select all only select all visible rows seems reasonable.

**Comment by jbetancur** [comment]
Yea, definitely select all will need to be disabled during paginationServer when using any of these 2 options.

Coverage is angry because it's falling below a threshold that I set. There are still some cases that are missing in tableReducer.

I'll try to get some time to review today or this week.

**Comment by jbetancur** [comment]
Ok, this looks good! Thank you! I'll merge this follow up with another PR to do the select All disabled and some minor tweaks.

**Comment by torregianif** [comment]
Hi guys,
Im having a problem with this case, i am using server pagination and selectableRowSelected and persistSelectedOnPageChange both in true but i am getting the gridData.selectedRows on onSelectedRowsChange with duplicates. 
For example: I select 2 rows from first page, then go to second page and back to first page and now i get 4 selected which are both rows of the first page duplicated. It seems that this piece of code isn't working: 

```
 case 'SELECT_MULTIPLE_ROWS': {
      const { selectedRows, rows, mergeSelections } = action;
      const selections = !mergeSelections ? selectedRows :
       **[...state.selectedRows, ...selectedRows.filter(row => !state.selectedRows.includes(row))];**
 return {
        ...state,
        selectedCount: selections.length,
        allSelected: selections.length === rows.length,
        selectedRows: selections,
      };
    }
```

Can you please check if these is only hapenning to me or is it actually a bug? In that case, do you have any ideas what i am doing wrong?

Thanks!

**Comment by soycoder** [comment]
> Hi guys, Im having a problem with this case, i am using server pagination and selectableRowSelected and persistSelectedOnPageChange both in true but i am getting the gridData.selectedRows on onSelectedRowsChange with duplicates. For example: I select 2 rows from first page, then go to second page and back to first page and now i get 4 selected which are both rows of the first page duplicated. It seems that this piece of code isn't working:
> 
> ```
>  case 'SELECT_MULTIPLE_ROWS': {
>       const { selectedRows, rows, mergeSelections } = action;
>       const selections = !mergeSelections ? selectedRows :
>        **[...state.selectedRows, ...selectedRows.filter(row => !state.selectedRows.includes(row))];**
>  return {
>         ...state,
>         selectedCount: selections.length,
>         allSelected: selections.length === rows.length,
>         selectedRows: selections,
>       };
>     }
> ```
> 
> Can you please check if these is only hapenning to me or is it actually a bug? In that case, do you have any ideas what i am doing wrong?
> 
> Thanks!

Hi
I found this happen too. Can anyone explain how to solve this? 

Thank you in advance.

**Comment by prasannakulal16** [comment]
Hii @jbetancur 
```
paginationServerOptions={{
            persistSelectedOnPageChange: true,
            persistSelectedOnSort: false
          }}
```
When i use above options 'Select All' option is not visible. Is there any solution to enable the 'Select ALL' ?

Issue #532, opened 2020-04-11
detectRTL method breaks when server side rendering


erikt9:
The `detectRTL` method uses `document` which will not be available when rendering in a server side environment.  Probably allowing the user to specify RTL as a prop as suggested by #514 and skip the `document` check if the prop is present is the best solution.  
My use case is enabling pagination which breaks when server side rendered because it always calls the detectRTL method (both with / without the paginationServer property).

jbetancur:
Sure, since there is a duplicate can you close this and just comment on the existing issue?

jbetancur:
implemented in 6.6.0

Issue #514, opened 2020-04-05
Support rtl from prop


net-bet:
Hi,

The lib determines rtl, via searching a root with an rtl direction settings.
But sometimes you can change direction through the DOM so this behavior might not be always accurate.
Is there a possibility to option to pass the direction direcly via rtl or direction prop on the table?
It will help for whoever uses a direction context, so he can jast map the direction context value to this prop and keep the behavior he wants 

I also saw this approach in other libs in react.

jbetancur:
This will be available in 6.6.0

PR #536, opened 2020-04-12
implement force rtl

jbetancur:
Implements #514 and #532

### DIFF
--- a/src/DataTable/Pagination.js
+++ b/src/DataTable/Pagination.js
@@ -62,6 +62,7 @@ const Pagination = ({
 }) => {
   const {
     data,
+    rtl,
     paginationRowsPerPageOptions,
     paginationIconLastPage,
     paginationIconFirstPage,
@@ -69,6 +70,7 @@ const Pagination = ({
     paginationIconPrevious,
     paginationComponentOptions,
   } = useTableContext();
+  const isRTL = rtl || detectRTL();
   const numPages = getNumberOfPages(rowCount, rowsPerPage);
   const lastIndex = currentPage * rowsPerPage;
   const firstIndex = (lastIndex - rowsPerPage) + 1;
@@ -84,7 +86,7 @@ const Pagination = ({
   const handleFirst = useCallback(() => onChangePage(1), [onChangePage]);
   const handleLast = useCallback(() => onChangePage(getNumberOfPages(rowCount, rowsPerPage)), [onChangePage, rowCount, rowsPerPage]);
   const handleRowsPerPage = useCallback(({ target }) => onChangeRowsPerPage(Number(target.value), currentPage), [currentPage, onChangeRowsPerPage]);
-  const isRTL = detectRTL();
+
   const selectOptions = paginationRowsPerPageOptions.map(num => (
     <option
       key={num}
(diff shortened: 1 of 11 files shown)


src/DataTable/Pagination.js
```diff
@@ -62,13 +62,15 @@ const Pagination = ({
...
     data,
+    rtl,
     paginationRowsPerPageOptions,
     paginationIconLastPage,
     paginationIconFirstPage,
     paginationIconNext,
     paginationIconPrevious,
     paginationComponentOptions,
   } = useTableContext();
+  const isRTL = rtl || detectRTL();
```
erikt9, line 73: @jbetancur This check should be against `undefined`, otherwise even when explicitly specifying rtl of `false`, `detectRTL()` is still called.

jbetancur: The goal here was to keep the old behavior and allow the detection to still happen but to have the option to `rtl=true` to force rtl on always. 

`rtl=false` means let RDT happen naturally as it should and thus let it fallback to detectRTL, detectRTL is just a helper for flipping around the pagination controls for rtl layouts, nothing more.

In hindsight, the prop name rtl may be confusing when I should have named it `forceRTL`, but I don't feel like doing a major release to fix just to rename it

erikt9: There really isn't any old behavior to preserve as this prop was just introduced.  If the prop is undefined (as it will be for any old code) detectRTL() is called just as before.

For server side rendering of a table with pagination there needs to be a way to skip the detectRTL() code and the document call within it.  Currently RTL is explicitly `false`, detectRTL() is still called.  Currently server side rendering only works if you also want RTL.

jbetancur: Yes, the old behavior was that detectRTL is always called. Also, rtl is false by default, not undefined unless you explicitly set it to that:
https://github.com/jbetancur/react-data-table-component/blob/b910a25c1f6ec58b6bec76efc90b3399bd0129c6/src/DataTable/propTypes.js#L243

Anyway, there are at least 2 options:
1. taking out detectRTL and if someone needs rtl they can just use the rtl option explicitly. Of course, I have to now increment a major version since this is a breaking change wrt Pagination/rtl
2. have the detectRTL check if we can use the DOM, eg:
   `typeof window !== 'undefined' && window.document && window.document.createElement`
   But I have no SSR environment to test that in atm, though I think this should work.

I didn't have SSR use cases in mind when I built RDT, so I am open to suggestions so feel free to propose another solution or submit a PR.

jbetancur: option 2 for example:

```js
export const detectRTL = () => {
  const canUse = !(typeof window !== 'undefined' && window.document && window.document.createElement);

  if (canUse) {
    return document.getElementsByTagName('BODY')[0] === 'rtl' || document.getElementsByTagName('HTML')[0].dir === 'rtl';
  }

  return false;
};
```

erikt9: Yes, that would work for me and if it's the way you want to go I'm fine with it.  I just think a better API would be to have a property that can have the values true / false / undefined (or perhaps some kind of enum) mapping to RTL, LTR, autodetect.

jbetancur: Ok, I think the right API is somewhere in the middle but we should just use the same values defined by the direction spec:

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir

I'm going to release a 6.7.0 shortly with this and just pull the rtl prop and call it out in the Release notes as it was not well-thought out. 

@net-bet just letting you know to hold off on 6.6.0 and wait for 6.7.0.

jbetancur: 6.7.0 is now available

PR #553, opened 2020-04-24
hint: maxWidth under 100px won't work by itself

cliffordp:
As evidenced by https://codesandbox.io/s/react-data-table-cannot-set-maxwidth-less-than-100px-without-also-minwidth-uj5kp?file=/src/index.js

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ Nothing new here - we are using an array of object literals and properties to de
 | grow     | number | no       | [flex-grow](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-grow) of the column. This is useful if you want a column to take up more width than its relatives (without having to set widths explicitly).  this will be affected by other columns where you have explicitly set widths |
 | width    | string | no       | give the column a fixed width                                                                                 |
 | minWidth | string | no       | give the column a minWidth                                                                                    |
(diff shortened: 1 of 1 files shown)


cliffordp:
@jbetancur actually, in testing this with _multiple 20px columns in a row,_ the column titles imploded/jumbled up on top of each other. Not sure if that's a bug to be fixed or if **minWidth** (in addition to this PR) should state that it shouldn't ever be less than 100px or bad things could happen.

jbetancur:
@cliffordp This is a challenge in building a table library, keeping some control over the min-width of columns while not allow things to break when people try to break things 😉 . I think it's good to call it out on the README so thanks for the PR! 😄 

I think we can work it a litle differently:

jbetancur:
I think we can work it a little differently to some variation of:

*Note: minWidth defaults to 100px. If you require a column smaller than 100px `minWidth` can be overridden using a static `width` or setting a `minWidth` lower than 100px*

cliffordp:
@jbetancur which attribute do you want that comment on - maxWidth?

cliffordp:
@jbetancur please reply to my prior comment or confirm that closing this is your preference.

jbetancur:
Sorry @cliffordp I must have missed this. `minWidth could be re-written as it overrides the default 100px if you want to resubmit a PR

PR #558, opened 2020-04-27
custom sortFunction for columns

### LINKED ISSUE #556: [FEATURE]: sortFunction for columns
## Is your feature request related to a problem? Please describe

When rendering custom cells in a data table, it is useful to set a custom sorting function for the column as well. Right now, we can set `sortFunction` for whole data table (as seen in [docs](https://www.npmjs.com/package/react-data-table-component#sorting)). But that means we have to implement sorting function for all columns if at least one of them is customized.

```
const columns = [
    {
        name: 'MyValue',
        selector: 'mySelector',
        cell: item => 'imagine some complex cell rendering here ' + item.toString()
    }, {
        name: 'Year',
        selector: 'year',
    }
]


const data = [{myIndex: 1, year: 2019}, {myIndex: 2, year: 2020}, ...]
const sortFunction = (rows, field, direction) => {
    if (field == 'mySelector') {
        // ... sort rows by mySelector ...
    } else {
        // ... sort for regular columns
    }
}
<DataTableComponent columns={columns} data={data} sortFunction={sortFunction}/>
```

## Describe the solution you'd like
We can add new column attribute `sortFunction` that accepts same `compareFunction` function that is used in JS [sort](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort)): `<T>(a: T, b: T): 1 | 0 | -1`

```
const columns = [
    {
        name: 'MyValue',
        selector: 'mySelector',
        cell: item => 'image some complex cell rendering here ' + item.toString()
        sortFunction: (a, b) => a.myIndex - b.myIndex,
    }
]


const data = [{myIndex: 1}, {myIndex: 2}, ...]
<DataTableComponent columns={columns} data={data} />
```

ondratra:
Fixes #556


jbetancur:
Thanks for the pr! This is an interesting feature. I’ll have a look at it this week

src/DataTable/DataTable.js
```diff
@@ -189,12 +189,23 @@ const DataTable = memo(({
 
   const sortedData = useMemo(() => {
     // server-side sorting bypasses internal sorting
-    if (!sortServer) {
+    if (sortServer) {
+      return data;
+    }
+
+    // use general sorting function when columns has no sort function on it's own
+    const column = sortColumn && columns.find(item => item.selector === sortColumn);
```
jbetancur, line 197: @ondratra sorry it took me so long to review this!

Overall this looks good, but I am not sure about adding an O(n) lookup on every sort (when `sortDirection` changes). Perhaps there is another way we can refactor this and avoid this extra step? I realize that there are finite columns here and I want to be careful of pre-optimization (perhaps we can measure this to be sure), but sorting is already pretty expensive and I am sensitive to performance issues.

src/DataTable/DataTable.js
```diff
@@ -189,12 +189,23 @@ const DataTable = memo(({
...
     }
 
-    return data;
-  }, [data, sortColumn, sortDirection, sortFunction, sortServer]);
+    // use column's custom sorting function
+    const customSortFunction = sortDirection === 'asc'
+      ? column.sortFunction
+      : (a, b) => column.sortFunction(a, b) * -1;
+
+    return [...data].sort(customSortFunction);
```
jbetancur, line 207: I like that we are using native sort!

Side note: I'd love to get away from lodash, unfortunately it has prop lookups that are quite useful so this is a good start.

jbetancur:
@ondratra just one comment to address around sort performance. Perhaps this is something we can measure to make sure we haven't impacted sorting performance

ondratra:
@jbetancur Hi :) I somehow missed your reply to this PR. 

I like your performance orientation and find it reasonable. I added a dictionary (key-value) of columns that removes the need for walking through columns during sorting.

jbetancur:
merged but something is busted with linting. I will push a PR with fixes

Issue #582, opened 2020-05-23
Expandable button position: Being able to have the expandable button at the end of the row (last column)


DiegoVargasGAP:
## Feature Check list

- [x] Agree to the [Code of Conduct](https://github.com/jbetancur/react-data-table-component/blob/master/CODE-OF-CONDUCT.md)
- [x] Read the README to ensure the feature is not already present
- [x] You read [Creating Issues, Features and Pull Requests](https://github.com/jbetancur/react-data-table-component/issues/387)
- [x] Considered the value versus complexity for all users of the library as well as library maintenance
- [x] Considered if this can be a storybook or documentation example

## Is your feature request related to a problem? Please describe

We want to have the expandable button in the last column

## Describe the solution you'd like

to have a property like expandableRowsLast so you can choose the position of the expandable button being it the first or last column

PR #583, opened 2020-05-23
Property for being able to have expandable buttons column as last column

DiegoVargasGAP:
Added property to data table and data row to be able to have the expandable buttons column as the last column of the grid, you can still have it in the first column if the property is not specified
Tests ran successfully #582


jbetancur:
Can you add an example to the Kitchen sink storybook? This also needs a unit test. A snapshot test should be ok.

Coder-SpiderMan:
@DiegoVargasGAP @jbetancur ,  Any update on this feature to add expandable buttons column as last column?

PR #587, opened 2020-05-27
Changed Pagination to use totalRows instead of data.length

jpuckety:
When using server-side pagination, the data.length is the length of one page and not appropriate for  calculating the value of 'All' in the paginator.


jbetancur:
Nice catch!

PR #590, opened 2020-05-29
Added option to reset sort column and direction to defaults

### LINKED ISSUE #570: How to reset the sort column & direction on some external event to DataTable?
## Issue Check list
- [x] Agree to the [Code of Conduct](https://github.com/jbetancur/react-data-table-component/blob/master/CODE-OF-CONDUCT.md)
- [x] Read the README
- [x] You are using React 16.8.0+
- [x] You installed `styled-components`
- [x] Include relevant code or preferably a [code sandbox](https://codesandbox.io/embed/react-data-table-sandbox-ccyuu
)

## Description
Update of `defaultSortField` & `defaultSortAsc` props doesn't change the sort column & direction.

## Relevant Code

```
// Table.jsx

import React from "react";
import DataTable from "react-data-table-component";
import { connect } from "react-redux";
import TitleCell from "./TitleCell";

const columns = [
{
    id: "title",
    name: "Title",
    sortable: true,
    right: true,
    cell: row => <TitleCell row={row} />
},
{
    // ... more sortable columns
}
];

const Table = props => {
    const { data, sortColumn, sortDirection, onSort } = props;
    const handleSort = (column, direction) => onSort(column.id, direction);

    return(
        <DataTable
            data={data}
            columns={columns}
            onSort={handleSort}
            sortFunction={r => r}
            defaultSortField={sortColumn}
            defaultSortDirection={sortDirection}
    );
};

export default connect(
    state => {
        return {
            data: getTableData(state),
            sortColumn: getSortColumn(state),
            sortDirection: getSortDirection(state)
        };
    },
    dispatch => {
        onSort: (sortColumn, sortDirection) => dispatch(Actions.changeSort(sortColumn, sortDirection)
    }
)(Table);

// tableReducer.jsx

const initialState = {
    sortColumn: "some_default_col",
    sortDirection: "asc"
};

export const tableReducer = (state = initialState, action) => {
    switch(action.type) {
        case "CHANGE_SORT":
            return { ...state, sortColumn: action.sortColumn, sortDirection: action.sortDirection };
    }
}
```

## Steps to reproduce
* Click on any other column than the default in the table header to update the sort.
* On some external event (such as button click), reset the table reducer's `sortColumn` & `sortDirection` to defaults again. (this will update the Table component props, but DataTable's sort remains the same)

## Expected behavior
Update in the `defaultSortField` & `defaultSortDirection` props should rerender the datatable to the updated sort column & direction.

## Additional context
Like `paginationResetDefaultPage`, a prop like `sortResetDefault` can solve the issue.

sk364:
#### Changes:
* Added prop `sortResetDefaultField` which when true resets `sortColumn` & `sortDirection` with respect to `defaultSortField` & `defaultSortDirection`

Closes #570

### DIFF
--- a/src/DataTable/DataTable.js
+++ b/src/DataTable/DataTable.js
@@ -143,17 +143,6 @@ const DataTable = memo(({
     visibleOnly: selectableRowsVisibleOnly,
     persistSelectedOnPageChange,
   });
-  const handleSortChange = (column, columnDirection) => dispatch({
-    type: 'SORT_CHANGE',
-    sortDirection: columnDirection,
-    sortColumn: column,
-    sortServer,
-    selectedColumn: {},
-    pagination,
-    paginationServer,
-    visibleOnly: selectableRowsVisibleOnly,
-    persistSelectedOnSort,
-  });
 
   useDidUpdateEffect(() => {
     onSelectedRowsChange({ allSelected, selectedCount, selectedRows });
@@ -178,10 +167,6 @@ const DataTable = memo(({
   useDidUpdateEffect(() => {
     handleChangePage(paginationDefaultPage);
   }, [paginationDefaultPage, paginationResetDefaultPage]);
-
-  useDidUpdateEffect(() => {
-    handleSortChange(defaultSortField, getSortDirection(defaultSortAsc));
-  }, [defaultSortField, defaultSortAsc, sortResetDefaultField]);
 
   useEffect(() => {
     if (selectableRowSelected) {
@@ -293,6 +278,9 @@ const DataTable = memo(({
     paginationComponentOptions,
     direction,
     mergeSelections,
+    defaultSortField,
+    defaultSortDirection: getSortDirection(defaultSortAsc),
+    sortResetDefaultField,
   };
 
   const showTableHead = () => {
--- a/src/DataTable/DataTable.js
+++ b/src/DataTable/DataTable.js
@@ -88,6 +88,7 @@ const DataTable = memo(({
   sortIcon,
   onSort,
   sortFunction,
+  sortResetDefaultField,
   sortServer,
   expandableRowsComponent,
   expandableRowDisabled,
@@ -142,6 +143,17 @@ const DataTable = memo(({
     visibleOnly: selectableRowsVisibleOnly,
     persistSelectedOnPageChange,
   });
+  const handleSortChange = (column, columnDirection) => dispatch({
+    type: 'SORT_CHANGE',
+    sortDirection: columnDirection,
+    sortColumn: column,
+    sortServer,
+    selectedColumn: {},
+    pagination,
+    paginationServer,
+    visibleOnly: selectableRowsVisibleOnly,
+    persistSelectedOnSort,
+  });
 
   useDidUpdateEffect(() => {
     onSelectedRowsChange({ allSelected, selectedCount, selectedRows });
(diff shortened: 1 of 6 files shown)


sk364:
Ping! @jbetancur

jbetancur:
Hey, @sk364, I'll take a look this week.

jbetancur:
@sk364 Thanks for the PR! So, pulled your changes locally and tried setting `sortResetDefaultField`. Can you tell me how this is used? Are you resetting this prop from the parent component to trigger the re-render?

It would be preferable to handle this without the extra prop and have the sorting behavior reset when a new column is hovered/selected by default. However, this is a bit more involved as it requires some refactoring to sorting.

Material-ui table example demonstrates the exact behavior that I think RDT could have:

https://codesandbox.io/s/f71wj

1. let the user sort the default directions across all columns.
2. if a field is sorted in descending then the next col to hover over should show the ascending arrow

sk364:
> @sk364 Thanks for the PR! So, pulled your changes locally and tried setting `sortResetDefaultField`. Can you tell me how this is used? Are you resetting this prop from the parent component to trigger the re-render?

the option works the same way as `paginationResetDefaultPage`. Parent component will switch it to true (based on some external event to data table) to reset to default field. Not sure how this will be handled without the extra prop.

jbetancur:
@sk364 I still think a better approach would be to get the sorting behavior fixed internally so that after sorting a column that when you hover on another column and then click it, it should respect the `defaultSortAsc` prop. This way we avoid changing the API while keeping the sorting config simple.

sk364:
@jbetancur but how would we reset sort field and direction to defaults when we update the table data?

sk364:
@jbetancur I think the "hovering on another column and then clicking it" part can be taken up in conjunction to the issue resolved by the PR. Do you want me to add it in this PR itself?

jbetancur:
Hey @sk364 apologies but in my attempt to multitask I conflating sorting behavior with the intent of your PR 😬 which is to just toggle the sorting to `defaultSortAsc`

Please see my comments, but the gist is instead of adding another handle sort in DataTable we should pass it down to Table Col via context and handle the effect there.

Also, you'll want to update the README with the new prop in the appropriate section

src/DataTable/DataTable.js
```diff
@@ -142,6 +143,17 @@ const DataTable = memo(({
     visibleOnly: selectableRowsVisibleOnly,
     persistSelectedOnPageChange,
   });
+  const handleSortChange = (column, columnDirection) => dispatch({
```
jbetancur, line 146: Instead of calling sort in a different place than TableCol can't you just pass the `sortResetDefaultField` via context to TableCol and handle the effect there. You can re-use the handleSort logic there

jbetancur:
I should have asked before but what is the use case for this? Why do we need to reset sorting? I can't find another table lib that does this.

sk364:
> I should have asked before but what is the use case for this? Why do we need to reset sorting? I can't find another table lib that does this.

In the app, the sort field and direction are stored in a reducer (just like DataTable). A custom sort method consumes this redux state.
There is some button which when clicked fetches more data, upon which the app's reducer resets the sort state, but DataTable doesn't.
Basically, the app wants the DataTable to reset completely whenever the user clicks that specific button.

Maybe a global reset for the DataTable would be a better option for this use case, but for me, pagination and sort resets are good enough for now.

sk364:
@jbetancur added the `sortResetDefaultField` in `DataTableContext`. The context is not refreshing when I am toggling the prop. Am I missing something here?

codytodonnell:
@jbetancur are there any plans to implement your proposed change regarding the default sort direction when sorting on a new column (i.e. replicate the same sorting behavior as Material-ui table)?

marcoangelo:
@jbetancur bumping this. Could really use a feature like this. Any plans to implement it ? thanks

PR #594, opened 2020-06-03
Fix wrong property type in README documentation

chucklam:
`subHeader` is a boolean for whether to show a subhead or not. It's not a component, which is what `subHeaderComponent` is.


jbetancur:
Nice catch. Thank you!

PR #622, opened 2020-07-07
Improve typing for onRowClicked and onRowDoubleClicked props

hajnalbendeguz:
A mouse event is passed as a 2nd paramater both times, but they were missing from the interface declaration.


hajnalbendeguz:
Found another one, it is ignoreRowClick, not ignoreOnRowClick.

jbetancur:
Nice catch and thank you!

PR #641, opened 2020-07-31
Allow pagination to be at the top instead of the bottom

alecf:
This makes it easy to add pagination at the top. 

I wasn't sure if this should go before/after TableSubheader, or inside of it.

I can update the docs in the PR if the general approach is acceptable.

### DIFF
--- a/src/DataTable/DataTable.js
+++ b/src/DataTable/DataTable.js
@@ -294,6 +294,17 @@ const DataTable = memo(({
 
   const showSelectAll = persistSelectedOnPageChange || selectableRowsNoSelectAll;
 
+  const renderPagination = () => (
+    <Pagination
+      onChangePage={handleChangePage}
+      onChangeRowsPerPage={handleChangeRowsPerPage}
+      // eslint-disable-next-line react/prop-types
+      rowCount={paginationTotalRows || data.length}
+      currentPage={currentPage}
+      rowsPerPage={rowsPerPage}
+    />
+  );
+
   return (
     <ThemeProvider theme={currentTheme}>
       <DataTableProvider initialState={init}>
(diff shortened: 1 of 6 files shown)


src/DataTable/DataTable.js
```diff
@@ -312,6 +313,16 @@ const DataTable = memo(({
             />
           )}
 
+          {enabledPagination && paginationPosition === 'top' && (
+            <Pagination
```
jbetancur, line 317: Rather than repeat the pagination code it might be better to have this as a separate code function that returns the pagination component so {enabledPagination && paginationPosition === 'top' && renderPagination()};

src/DataTable/DataTable.js
```diff
@@ -422,7 +433,7 @@ const DataTable = memo(({
             </Table>
           </TableWrapper>
 
-          {enabledPagination && (
+          {enabledPagination && paginationPosition === 'bottom' && (
```
jbetancur, line 436: same here but  {enabledPagination && paginationPosition === 'bottom' && renderPagination()};

src/DataTable/propTypes.js
```diff
@@ -125,6 +125,7 @@ export const propTypes = {
   paginationTotalRows: PropTypes.number,
   paginationPerPage: PropTypes.number,
   paginationRowsPerPageOptions: PropTypes.array,
+  paginationPosition: PropTypes.oneOf(['bottom', 'top']),
```
jbetancur, line 128: We should consider in a future version `both` so pagination renders on top and bottom.  I am not sure how this would look but this PR is a good setup for that later 😉

jbetancur:
@alecf Thanks for the PR! Just a few comments to clean things up

src/DataTable/DataTable.js
alecf: good idea, done!

src/DataTable/DataTable.js
alecf: done

alecf:
@jbetancur friendly ping! (may not have been obvious that I fixed the issues you brought up and re-requested the review)

src/DataTable/DataTable.js
```diff
@@ -312,6 +324,9 @@ const DataTable = memo(({
             />
           )}
 
+          {enabledPagination && paginationPosition === 'top' &&
```
jbetancur, line 327: On second look I thnk this needs to be below Table Subheader

alecf:
oh good point - I'll give it a try with the CSS switch

ayub-786 (commented):
i m also using this library and I m not getting how it's work for next js so please can you give me several example for this

ayub-786 (commented):
can you add this props parameter inside of react-data-table-component so it's helpful in the next js? I m not getting that props parameter in this library, so can you suggest how to show the pagination position on both sides of the table (top/bottom).

PR #665, opened 2020-08-25
Fix generic types

julioxavierr:
Having the generic types is really useful for inference and works really well for us.
I added a default to them, making it possible to be able to set them optionally.


jbetancur:
@julioxavierr I appreciate the fix. I did not realize the impact of removing these (mainly because I'm new to typescript). 

I'll just need a bit of time to review and test since it's a busy week.

On a side note, I am in the works of moving the entire code base to typescript

JamiesWhiteShirt (approved):

Issue #612, opened 2020-06-24
[FEATURE]: Option not to disable TableHeadRow when there's no data


philipnery:
## Feature Check list

- [X] Agree to the [Code of Conduct](https://github.com/jbetancur/react-data-table-component/blob/master/CODE-OF-CONDUCT.md)
- [X] Read the README to ensure the feature is not already present
- [X] You read [Creating Issues, Features and Pull Requests](https://github.com/jbetancur/react-data-table-component/issues/387)
- [X] Considered the value versus complexity for all users of the library as well as library maintenance
- [X] Considered if this can be a storybook or documentation example

## Is your feature request related to a problem? Please describe

We have buttons placed inside table row headers. 

Unfortunately, if the table has no data (and we're using `persistTableHead` BTW), it gets disabled so we can't anymore click the buttons inside.

## Describe the solution you'd like

Either we add another props for this or modify `TableHeadRow`'s `disabled` when `persistTableHead` is true.

Maybe something like this?

```javascript
                  <TableHeadRow
                    className="rdt_TableHeadRow"
                    role="row"
                    dense={dense}
                    disabled={progressPending || (data.length === 0 && !persistTableHead)}
                  >
```

##
Thanks,
Philip

jbetancur:
Thanks for raising this. I've not has as much time as I'd like to work but I will try to get to it the next week. PR's are welcome if you'd like to contribute and get a fix in sooner.

jbetancur:
I had a chance to take a lok. Can you share code or preferable a codesandbox https://codesandbox.io/embed/react-data-table-sandbox-ccyuu of how you are adding elements to TableHeadRow? 

TableHeadRow is meant to be internally controlled.

In any case, I think the disabled can be moved to `TableCol` `ColumnSortable` since it is the sort button that should be disabled or hidden if there is nothing to sort.

The

mathieupetrini:
Thanks for your library.
Any updates on this ?

jbetancur:
Implemented this in RDT `next` branch/alpha-6.
https://github.com/jbetancur/react-data-table-component/pull/776

PR #699, opened 2020-10-13
header clickable even if no data

mathieupetrini:
Related to #612


mathieupetrini:
@jbetancur Can you take a look to this PR ?

jbetancur:
@mathieupetrini thanks for the PR. Sorry for the delay as I have been quite busy.  So are you proposing to remove disabling the header and just disable to individual TableCols? Curious if there a specific use case that led to this?

mathieupetrini:
The goal of this PR is to remove disabling the header when there is any data.

The purpose is to keep filter include in header available to be changed.

src/DataTable/DataTable.js
```diff
@@ -335,7 +335,6 @@ const DataTable = memo(({
                     className="rdt_TableHeadRow"
                     role="row"
                     dense={dense}
-                    disabled={progressPending || data.length === 0}
```
jbetancur, line 338: Can you also remove the css check in https://github.com/jbetancur/react-data-table-component/blob/6e9d9246076f9b39663c942fc69848ec6744edde/src/DataTable/TableHeadRow.js#L13

src/DataTable/DataTable.js
```diff
@@ -350,6 +349,7 @@ const DataTable = memo(({
                         key={column.id}
                         column={column}
                         sortIcon={sortIcon}
+                        sortable={!progressPending && data.length > 0}
```
jbetancur, line 352: this is a little confusing when we also have `column.sortable` how about a rename this to `disabled`?

src/DataTable/TableCol.js
```diff
@@ -15,12 +15,12 @@ const ColumnSortable = styled.div`
   height: 100%;
   line-height: 1;
   user-select: none;
-  ${props => (props.sortActive ? props.theme.headCells.activeSortStyle : props.theme.headCells.inactiveSortStyle)};
+  ${props => (props.sortActive && props.sortable ? props.theme.headCells.activeSortStyle : props.theme.headCells.inactiveSortStyle)};
```
jbetancur, line 18: props.disabled

src/DataTable/TableCol.js
```diff
@@ -15,12 +15,12 @@ const ColumnSortable = styled.div`
...
 
   span.__rdt_custom_sort_icon__ {
     i,
     svg {
-      ${props => (props.sortActive ? 'opacity: 1' : 'opacity: 0')};
+      ${props => (props.sortActive && props.sortable ? 'opacity: 1' : 'opacity: 0')};
```
jbetancur, line 23: props.disabled

src/DataTable/TableCol.js
```diff
@@ -120,18 +121,18 @@ const TableCol = memo(({
           role="columnheader"
           tabIndex={0}
           className="rdt_TableCol_Sortable"
-          onClick={handleSortChange}
-          onKeyPress={handleKeyPress}
-          sortActive={sortActive}
+          onClick={sortable ? handleSortChange : undefined}
```
jbetancur, line 124: `disabled`

src/DataTable/TableCol.js
```diff
@@ -144,6 +145,11 @@ TableCol.propTypes = {
     PropTypes.bool,
     PropTypes.object,
   ]).isRequired,
+  sortable: PropTypes.bool,
+};
+
+TableCol.defaultProps = {
+  sortable: true,
```
jbetancur, line 152: `disabled`

src/DataTable/TableCol.js
```diff
@@ -52,6 +52,7 @@ const ColumnSortable = styled.div`
 const TableCol = memo(({
```
jbetancur, line 52: remove the extra line space above

jbetancur (changes requested):
A few comments on naming. Also, we also need a snapshot test for this case scenario. Give the test a shot otherwise I can add it later.

mathieupetrini:
I had a snapshot test. I'm not sure about is quality if you can give a look.

Tanks for reviewing

src/DataTable/TableHeadRow.js
```diff
@@ -1,16 +1,12 @@
...
+import styled from 'styled-components';
 
 const TableHeadRow = styled.div`
   display: flex;
   align-items: stretch;
   width: 100%;
   ${props => props.theme.headRow.style};
   ${props => (props.dense && props.theme.headRow.denseStyle)};
-  ${props => props.disabled && disabled};
+  ${props => props.disabled};
```
jbetancur, line 9: This line should be removed

jbetancur (changes requested):

src/DataTable/__tests__/DataTable.test.js
```diff
@@ -2794,4 +2794,26 @@ describe('DataTable::direction', () => {
 
     expect(container.firstChild).toMatchSnapshot();
   });
+
+  test('should render correctly when sortable column is false', () => {
+    let testClicked = false;
+    const mock = dataMock({
+      // eslint-disable-next-line jsx-a11y/no-static-element-interactions, jsx-a11y/click-events-have-key-events, no-return-assign
+      name: (<div id="testColName" onClick={() => testClicked = true}>Test</div>),
```
jbetancur, line 2802: take a look at the other tests but you can just use jest.fn and then test if it was called:

```js
    const testClicked = jest.fn();
    const mock = dataMock({
      // eslint-disable-next-line jsx-a11y/no-static-element-interactions, jsx-a11y/click-events-have-key-events, no-return-assign
      name: (<div id="testColName" onClick={testClicked}>Test</div>),
      sortable: false,
    });

    const { container } = render(
      <DataTable
        data={mock.data}
        columns={mock.columns}
        selectableRows
      />,
    );

    fireEvent.click(container.querySelector('#testColName'));

    expect(testClicked).toBeCalled();
    expect(container.firstChild).toMatchSnapshot();
```

src/DataTable/__tests__/DataTable.test.js
```diff
@@ -2794,4 +2794,26 @@ describe('DataTable::direction', () => {
 
     expect(container.firstChild).toMatchSnapshot();
   });
+
+  test('should render correctly when sortable column is false', () => {
```
jbetancur, line 2798: I'm not sure this test is correct. We want to setup a case on the table where the snapshot will reflect the state of the table when:

- persistTableHead is true (the headers only show up if this is true)
- progressPending is true
- you cannot sort ie sort is not called:

I would also move these tests to thhe line 492 describe block `describe('when persistTableHead', () => {`

for the test try something like:

```js
  test('should not call onSort if progressPending true', () => {
    const onSortMock = jest.fn();
    const mock = dataMock({ sortable: true });
    const { container } = render(
      <DataTable
        data={mock.data}
        columns={mock.columns}
        onSort={onSortMock}
        persistTableHead
       progressPending
      />,
    );

    fireEvent.click(container.querySelector('div[id="column-some.name"]'));

    expect(onSortMock).not.toBeCalled();
  });
```

jbetancur: Actually, with this test a snapshot is not needed

mathieupetrini: Sorry, the test was wrong indeed.

What we want to test is : persistTableHead and no data. We need to be able to click on header column if there is a function bind.
And the sort function must NOT be called, you are right.

src/DataTable/__tests__/DataTable.test.js
```diff
@@ -545,6 +545,31 @@ describe('DataTable::progress/nodata', () => {
...
       expect(container.firstChild).toMatchSnapshot();
     });
+
+    test('should render correctly when persistTableHead and no data', () => {
+      const functionHadToBeClicked = jest.fn(() => true);
+      const onSortMock = jest.fn();
+
+      const mock = dataMock({
+        // eslint-disable-next-line jsx-a11y/no-static-element-interactions, jsx-a11y/click-events-have-key-events
+        name: (<div id="testColName" onClick={functionHadToBeClicked}>Test</div>),
```
jbetancur, line 555: Why is the name block is needed or funcitonHadtoBeClicked? You just need to mock someone attempting to click the cell when data is 0 or prrogresspending is true

jbetancur (changes requested):

src/DataTable/__tests__/DataTable.test.js
jbetancur: these are the tests you need

```js
  test('should not call onSort if progressPending true', () => {
    const onSortMock = jest.fn();
    const mock = dataMock({ sortable: true });
    const { container } = render(
      <DataTable
        data={mock.data}
        columns={mock.columns}
        onSort={onSortMock}
        persistTableHead
        progressPending
      />,
    );

    fireEvent.click(container.querySelector('div[id="column-some.name"]'));

    expect(onSortMock).not.toBeCalled();
  });
```

and

```js
  test('should not call onSort if data is empty []', () => {
    const mock = dataMock({ sortable: true });
    const { container } = render(
      <DataTable
        data={[]}
        columns={mock.columns}
        onSort={onSortMock}
        persistTableHead
      />,
    );

    fireEvent.click(container.querySelector('div[id="column-some.name"]'));

    expect(onSortMock).not.toBeCalled();
  });
```

mathieupetrini: No, i even want to be able to click on div or focus input in TableCol.

For example, here i cannot click on select under "Status" ...
image

jonathanxlee:
Is there any update on this feature. This would be very helpful.

mounicagdv:
how to make the columns header clickable for sorting

abaum-augu:
Any updates on this? @jbetancur if you still interested on a use case for this, we are developing a multi table component to which you can pass single tables as child.
```
<MultiTable>
<Table>
<Table>
...
</MultiTable>
```
The Multi table has a header that controls sorting for all of the child tables. That header is actually an empty data table. So in this case the sorting buttons are disabled, therefore can't sort the children tables data.

Issue #704, opened 2020-10-16
[SUPPORT REQUEST] Sorting state (column, direction) does not persist


JaffParker:
## Issue Check list
- [x] Agree to the [Code of Conduct](https://github.com/jbetancur/react-data-table-component/blob/master/CODE-OF-CONDUCT.md)
- [x] Read the README
- [x] You are using React 16.8.0+
- [x] You installed `styled-components`
- [x] Include relevant code or preferably a [code sandbox](https://codesandbox.io/embed/react-data-table-sandbox-ccyuu
)

## Describe the bug
I was very much on the fence about opening this issue as it's very much a support request, but I'm out of options. I'm grateful for any help you'd be able to give.

I am creating a table with server-side pagination and sorting. No issues with pagination, however, sorting does not work as intended. When I click on a column to sort, it does pass the sorting parameters to my handler, but does not save the state. So if I go to click it a second time, it'll pass the exact same parameters (eg, id+asc).

## Expected behavior

Sorting order and column get persisted in the state.

## Code Sandbox, Screenshots, or Relevant Code

I was not able to reproduce it in a codesandbox, here's my attempt though: https://codesandbox.io/s/react-data-table-sandbox-forked-gu9rt. Might be because there's no actual server. Here's the code responsible for this table:

```ts
export const ListContainer: FC<ListContainerProps> = ({
  userId,
}) => {
  const [page, setPage] = useState(1)
  const [perPage, setPerPage] = useState(10)
  const [sort, setSort] = useState<StudyParticipantsSort>({
    id: SortOrder.ASC,
  })
  const { data, loading } = useListContainerQuery({
    variables: {
      userId,
      page: {
        count: perPage,
        page,
      },
      sort,
    },
    fetchPolicy: 'cache-and-network',
  })

  const handleRowsPerPageChange = useCallback(
    (currentRowsPerPage: number, currentPage: number) => {
      setPerPage(currentRowsPerPage)
      setPage(currentPage)
    },
    [],
  )

  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  const handleSort = useCallback((column: any, order: 'desc' | 'asc') => {
    setSort({ [column.id as string]: order.toUpperCase() })
  }, [])

  return (
    <DataTable
      progressPending={loading}
      progressComponent={<Loading />}
      pagination
      paginationServer
      paginationTotalRows={data?.participantsFeed.resourceInfo.totalCount}
      paginationPerPage={perPage}
[...]

PR #719, opened 2020-11-16
Switch sorting identifier to ID instead of selector

JaffParker:
In TableCol I switched it to identify sorted column by ID instead of selector, as per [this issue](https://github.com/jbetancur/react-data-table-component/issues/704).

Since selectors can often be functions, I also removed them from the ID of the column, substituting it for ID as well. That required me to make changes to the tests as well.

### DIFF
--- a/src/DataTable/TableCol.js
+++ b/src/DataTable/TableCol.js
@@ -63,14 +63,14 @@ const TableCol = memo(({
       let direction = sortDirection;
       // change sort direction only if sortColumn (currently selected column) is === the newly clicked column
       // otherwise, retain sort direction if the column is switched
-      if (sortColumn === column.selector) {
+      if (sortColumn === column.id) {
         direction = sortDirection === 'asc' ? 'desc' : 'asc';
       }
 
       dispatch({
         type: 'SORT_CHANGE',
         sortDirection: direction,
-        sortColumn: column.selector,
+        sortColumn: column.id,
         sortServer,
         selectedColumn: column,
         pagination,
@@ -101,7 +101,7 @@ const TableCol = memo(({
     </span>
   );
 
-  const sortActive = column.sortable && sortColumn === column.selector;
+  const sortActive = column.sortable && sortColumn === column.id;
   const nativeSortIconLeft = column.sortable && !sortIcon && !column.right;
   const nativeSortIconRight = column.sortable && !sortIcon && column.right;
   const customSortIconLeft = column.sortable && sortIcon && !column.right;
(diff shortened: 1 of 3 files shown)


jbetancur:
@JaffParker thank you for the PR! I'll review this shortly.  I think one case we need is a test if no column id is present since RDT will auto-generate column id's if they are missing

src/DataTable/TableCol.js
```diff
@@ -63,14 +63,14 @@ const TableCol = memo(({
...
-      if (sortColumn === column.selector) {
+      if (sortColumn === column.id) {
         direction = sortDirection === 'asc' ? 'desc' : 'asc';
       }
 
       dispatch({
         type: 'SORT_CHANGE',
         sortDirection: direction,
-        sortColumn: column.selector,
+        sortColumn: column.id,
```
jbetancur, line 73: Looks like this breaks sorting  because `sortColumn` is expected to be string based on the selector name.

jbetancur: `sortColumn` probably needs to be converted to the currently selected column object. It should probably also be renamed to be more clear

JaffParker: Right. I didn't test with built-in sorting, only server sorting. This action does have the entire column object in its payload.

I'd be a bit wary of major refactoring in a plain JS project. How interested would you be in switching to TS? Or at least introducing TS type checking into the code (I think it's possible to add it to JS code through comments)

jbetancur: I've had plans to convert the entire lib to TS - I just haven't had the time to dedicate to it.  Along with this would be a major refactor and possible breaking changes.

jbetancur: I think for now it's fine in the JS code as long as it's commented. As I mentioned it will eventually be TS as soon as I either get some cycles or get some help 😉

JaffParker: I don't mind helping you out with that. I'm looking to get more involved in open source anyway :) I work in TS regularly and your lib already has typings, that facilitates it.

But we can go with comments for now too. I haven't worked with them all that much, but it's not complicated. Would just get bulky, that's all.

jbetancur:
@JaffParker I am adding this fix to the next branch which is the next release of RDT in TypeScript. Fixing this is actually a blocker uncovered by TypeScript.  Because this is going to require a pretty extensive refactor, I am going to go a different route and will make the default sort field based on `column.id`. Unless there are better ideas out there 😉

PR #730, opened 2020-12-05
Added new prop persistCurrentPageOnSort for paginationServerOptions 

vijicode:
Added new prop persistCurrentPageOnSort for paginationServerOptions to keep currentPage on sort_change event.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -152,6 +152,7 @@ export interface IDataTableProps<T = any> {
   paginationServerOptions?: {
     persistSelectedOnSort?: boolean;
     persistSelectedOnPageChange?: boolean;
+    persistCurrentPageOnSort?: boolean;
   };
   paginationDefaultPage?: number;
   paginationResetDefaultPage?: boolean;
--- a/src/DataTable/DataTable.js
+++ b/src/DataTable/DataTable.js
@@ -125,7 +125,7 @@ const DataTable = memo(({
     selectedColumn,
     sortDirection,
   }, dispatch] = useReducer(tableReducer, initialState);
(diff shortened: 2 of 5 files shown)


vijicode:
Could you please take a look at this proposed change ? @jbetancur   Thank you so much.

ninjaengineers:
@jbetancur any update on this?

jbetancur:
Sorry for the delay, I’ve tagged this into the next version as I’m currently migrating the code base to typescript. So, I’ve been a bit focused on that. I need sometime to test this scenario and a see how it integrates with the future RDT design.

Can you step me through some scenarios as to what this changes? Are we sure this does not break anything pagination related?

vijicode:
I have tried to add a new prop called "persistCurrentPageOnSort" ( false by default). If the developer is implementing server-side pagination and No server-side sorting, then it means that the sorting will depend on the per-page sorting with whatever table data is on that particular page. Therefore , to reset the page number back to 1 will not be suitable. Hence I added a check, if pagination && paginationServer && persistCurrentPageOnSort then keep the page number same as the one in the current state.
Example scenario:
server-side-pagination - returns 5 pages
server-side-sorting - not implemented
UX - user navigates to page 4 and clicks sort button on any column
UX - user will stay on page 4, contents of page4 will be sorted 
( Without fix, user will get sent to page 1 )

Thanks.

emanuallan:
Was this integrated elsewhere because this would be useful!

AhmadRaza09:
Please merged this issue. this is helpful

AhmadRaza09:
Please merged this issue. this is helpful

@jbetancur

GregoryInnovo:
This change is very important in some use cases

Issue #422, opened 2020-01-10
enabling fixedHeader and having columns with dynamic width causes the table header row to be misaligned with the body rows


muubar:
First of all, thank you for the great effort, I've been using RDT at work for a while now, so again, thank you.

## Describe the bug
when I enable `fixedHeader` the table body is set to have `overflow-y: scroll`, which will always show the vertical scrollbar, this is a correct behavior, however, **this causes the content width of the header row to be larger than the content width of body rows**, this happens because the scrollbar is taking up space from the table body row as well.
this starts to be a problem if I have `grow: '1'` on some of my columns, which means we now have column's with dynamic width, but now **the width of the column head is no longer equal to the cells width under the same column**, because again, the overall width of the header row is larger than the width of the body row.
this whole issue can be simply solved by adding `padding-right: SCROLLBAR_WIDTH` to the header row.

## To Reproduce
1. add columns with `width` only, and others with `minWidth` and `grow`, and have `fixedHeader` enabled
2. fill the headers with any data, and do the same for the body
3. notice how the columns with dynamic width are not aligning correctly with the body below them 

## Expected behavior
I expect each column header to align correctly with the cells in the body that are under the same column 

## Code Sandbox
https://codesandbox.io/s/react-data-table-sandbox-2m8l9

### 1)
![row sample](https://i.postimg.cc/MHR4HPRs/row.png)
### 2)
![header sample with no padding](https://i.postimg.cc/KzCwZSKk/no-padding.png)
### 3)
![header sample with padding fix](https://i.postimg.cc/mrXK5FBS/with-padding.png)

## Versions
 - React: 16.8.6
 - Styled Components: 4.4.0
 - OS: windows 10
 - Browser: Chrome
 - Version: 79.0.3945.88

jbetancur:
Your welcome! I'll have a look at this. I will be completely honest in that the `fixedHeader` feature is not feature complete so it could use some TLC

muubar:
@jbetancur I think you might find using [this hook](https://github.com/streamich/react-use/blob/master/docs/useScrollbarWidth.md) useful if you decide to fix it the way I suggested, another solution I was thinking of is to wrap the table body inside a div, and let that div have a height that is equal to the table body's height, and then let this new div handle the overflowing content with a scrollbar, this way the scrollbar will be right outside the body of the table instead of inside of it; something like [this](http://jsfiddle.net/p75LE/2/)

jbetancur:
[...]

PR #737, opened 2020-12-23
Change table header to use sticky positioning

danspyer:
I have changed the table header to use the position: sticky attribute as per [this issue ](https://github.com/jbetancur/react-data-table-component/issues/422)

### DIFF
--- a/src/DataTable/DataTable.js
+++ b/src/DataTable/DataTable.js
@@ -328,7 +328,7 @@ const DataTable = memo(({
               </ProgressWrapper>
             )}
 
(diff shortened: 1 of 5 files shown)


jbetancur:
Hey @danspyer I appreciate the PR! I’ll be able to have a look at this later this week after xmas, but looks ok upon glance. Have you tested this across all major browsers? Sticky is sort of only recently supported. 

Also, just a note I am in the process of migrating rdt to typescript so if we merge this I’ll keep the issue open so we can include the change in rdt v7.

PR #766, opened 2021-01-30
added in a fixed footer

Avery246813579:
I added a way to create a fixed footer that will display as a fixed data row at the bottom of the table. This is really useful for tables that you want to have a total count at the bottom.

### DIFF
--- a/src/DataTable/DataTable.js
+++ b/src/DataTable/DataTable.js
@@ -424,6 +424,7 @@ const DataTable = memo(({
                   {footer && (
                     <TableRow
                       id={-1}
+                      rowIndex={-1}
                       keyField={keyField}
                       row={footer}
                       columns={columnsMemo}
--- a/src/DataTable/DataTable.js
+++ b/src/DataTable/DataTable.js
@@ -103,6 +103,7 @@ const DataTable = memo(({
   theme,
   customStyles,
   direction,
+  footer,
 }) => {
   const initialState = {
     allSelected: false,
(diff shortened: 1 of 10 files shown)


khwerhahn:
Is there some documentation about this feature? Couldn't find any.

Avery246813579:
> Is there some documentation about this feature? Couldn't find any.

This is a new feature so there is no documentation on master, but if you check the files changed you can see the documentation change in the README.

All you have to do is put row data into the footer prop and it will display that row data as a fixed footer.

nazimjamil:
Would really appreciate this feature -- PR looks good to me.

nazimjamil:
Also, @Avery246813579 doesn't look like you updated the types file for the footer, will need to add that before merging imo.

batamire:
Just started using react-data-table-component and immediately landed here! Showing footer/totals is really a basic feature for any table component :)

There's a mention of `.rdt_TableFooter` in CSS Overrides readme section but I fail to see where it is used...

batamire:
`footerBold` option breaks custom `cell: (props) => ...`

Avery246813579:
> > `footerBold` option breaks custom `cell: (props) => ...`
> 
> I've noticed that it turns the cell into a react component and to access the value, you have to get it from `props.children`. Odd, I can't quite work out why this is happening!

We surround the row with a <b> tag to force make it bold, if we just pass the props to the children it should fix. Give me a few hours and I'll look into this.

src/DataTable/DataTable.js
```diff
@@ -414,10 +416,46 @@ const DataTable = memo(({
...
                         selected={selected}
                         selectableRowsHighlight={selectableRowsHighlight}
+                        footer={false}
                         rowIndex={i}
                       />
                     );
                   })}
+
+                  {footer && (
+                    <TableRow
```
jbetancur, line 426: These are an awful lot of props that will never be used for a footer row.

jbetancur (commented):
Sorry, it's taken me some time to respond. Thank you for your PR.

Is the footer supposed to be sticky or just when you scroll down to the bottom? 
Or is the objective to have a footer as the last row of the table?

src/DataTable/DataTable.js
jbetancur: Perhaps a footer should be its own component `<Footer />` similar to `subHeaderComponent` Footer can maybe have access to some table state similar to `expandableRowsComponent`

jbetancur: For example, if aligning columns is not needed

https://github.com/jbetancur/react-data-table-component/blob/master/src/DataTable/ExpanderRow.js

this allows you to pass in `expandableRowsCompent={({ data }) => <div>{data.length}</div>}`

So let you can pass in whatever Footer you would like

`footerComponent={({ data, columns.... etc }) => <div>Count: {data.length}</div>}`

batamire: I guess all rows data to do calculations... Columns should be aligned with the ability to do colspans. And expandable seems like a nice option. Too much? 🤯

nazimjamil: I think a footer should be sticky, or at least be given the option to be, similar behaviour to the header, where it can be incorporated in the flow of the rows but also have the option to be fixed at the top so you can see what the row values relate to.

ajaykumar0787:
Hi @Avery246813579,

Is this available in the master yet? I tried installing the latest through npm and still cannot see a footer but I can see the updated readme though.

lucaspokaz:
I still couldn't put the footer as you mentioned. Can anyone provide an example?

I'm using 7.4.7

kenfoong1997:
Any update on this tq?

angelocala94:
I'm also waiting for this. Any update for a footer option?

Issue #586, opened 2020-05-27
Scroll bar takes up space in windows so cause disorientation of table


baymac:
## Issue Check list
- [ x] Agree to the [Code of Conduct](https://github.com/jbetancur/react-data-table-component/blob/master/CODE-OF-CONDUCT.md)
- [x ] Read the README
- [x ] You are using React 16.8.0+
- [ ] You installed `styled-components`
- [x ] Include relevant code or preferably a [code sandbox](https://codesandbox.io/embed/react-data-table-sandbox-ccyuu
)

## To Reproduce
Steps to reproduce the behaviour:
1. Access a windows system
2. Go to https://codesandbox.io/embed/react-data-table-sandbox-wf7n8?fontsize=14&hidenavigation=1&theme=dark
3. Scroll down in the table

## Expected behavior
In Mac the scrollbar doesn't take up space in the table data columns, but in windows it takes up space so the table is disoriented.

## Code Sandbox, Screenshots, or Relevant Code

https://codesandbox.io/embed/react-data-table-sandbox-wf7n8?fontsize=14&hidenavigation=1&theme=dark

On Windows:

<img width="777" alt="Screenshot 2020-05-27 at 11 44 14 AM" src="https://user-images.githubusercontent.com/23079344/82984292-7e2c2f80-a00f-11ea-921c-d59c14cedbb0.png">

On Mac:

<img width="777" alt="Screenshot 2020-05-27 at 11 45 26 AM" src="https://user-images.githubusercontent.com/23079344/82984352-a2880c00-a00f-11ea-8112-de5a5f5714d5.png">

## Versions (please complete the following information)
 - React (RDT requires 16.8.0+)
 - Styled Components
 - OS: Windows
 - Browser Chrome

## Additional context
Is there a way to override windows scrollbar and not make it behave like in mac?

jbetancur:
Thanks for including a code sandbox. This is an issue with the way `fixedHeader` is implemented and how the table is wrapped internally.  The short of it is that the way `fixedHeader` is designed needs to be re-thought. I think there are a couple open issues on this but I will flag this as an enhancement.

baymac:
@jbetancur Can we use 
```
::-webkit-scrollbar {
  display: none;
}
```
If yes what class should this be applied in order for data table (without header row) scrollbar is hidden?

TreyBastian:
[...]

PR #767, opened 2021-02-02
Fixed scrollbars appearing when they're not needed

Narcha:
When a table has `fixedHeader` enabled and not enough entries to fill it's `max-height`, a disabled scrollbar appears.
I fixed this by replacing `overflow-y: scroll` with `overflow-y: auto` so the scrollbar only appears when it's necessary.

I've tested a few other configurations to see if this change breaks anything,
I'm not aware of any negative impact this could have.
Please let me know if I missed anything.


jbetancur:
Thank you for PRing this. Sorry for the delay, but I still need some time to test it out. Also, we will need to PR this into the next branch as well.

jbetancur:
This looks good! I'll also cherry pick this change into the `next` branch as well

Narcha:
One thing to note:
If #586 gets fixed by somehow adding the correct amount of `padding-right` to `rdt_TableHeadRow` to compensate for the scrollbar width, this will have to be reevaluated.

jbetancur:
@Narcha I've moved #586 to `next` as well. I've been meaning to fix that issue for some time

PR #778, opened 2021-02-22
fix potential performance issue when using conditional rows/cells

### LINKED ISSUE #772: Conditional Row Styles create a new class instead of being inline style
## Issue Check list
- [ x] Agree to the [Code of Conduct](https://github.com/jbetancur/react-data-table-component/blob/master/CODE-OF-CONDUCT.md)
- [x ] Read the README
- [x ] You are using React 16.8.0+
- [x ] You installed `styled-components`
- [x ] Include relevant code or preferably a [code sandbox](https://codesandbox.io/embed/react-data-table-sandbox-ccyuu
)

## Describe the bug
Conditional Row Styles create a new class instead of being inline change and we get the following message:
```
Over 200 classes were generated for component styled.div with the id of "sc-jrAGrp".
Consider using the attrs method, together with a style object for frequently changed styles.
Example:
  const Component = styled.div.attrs(props => ({
    style: {
      background: props.background,
    },
  }))`width: 100%;`
```

  <Component />
## To Reproduce
Simply add Conditional row style that depends on row value to create css style and update the row frequently (for example the table updates when new data come from websocket).

```
    {
      when: _row => true,
      style: row => ({
        background: `linear-gradient(90deg, rgba(255,88,87,0.1) ${calculateBackgroundBar(row, BidsData)}, rgba(0,0,0,1) ${calculateBackgroundBar(row, BidsData)})`,
        color: 'white'
      }),
    }

```
## Expected behavior
styled-components shouldn't create over 200 classes

## Versions (please complete the following information)
 - React 17.0.1
 - Styled Components 5.2.1
 - OS:  Windows 10 2004
 - Browser: Chrome

jbetancur:
FIxes #772

### DIFF
--- a/src/DataTable/TableCell.tsx
+++ b/src/DataTable/TableCell.tsx
@@ -8,7 +8,6 @@ interface TableCellStyleProps {
 	wrapCell: boolean | undefined;
 	allowOverflow: boolean | undefined;
 	cellStyle: CSSObject | undefined;
-	extendedCellStyle: CSSObject;
 }
 
 const overflowCSS = css<TableCellStyleProps>`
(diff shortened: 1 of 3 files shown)


mbpreble:
Apologies for commenting on an old pull request. This change seems to make `conditionalRowStyles` much less flexible, as `TableRowStyle` injects any styles that can be inlined into the `style` attribute of the `div` element it renders, but no longer makes any modifications to the component's stylesheet.

Styling through `conditionalRowStyles` is therefore limited to a subset of what is passed in (something like the `CSSProperties` type in styled-components but not the more flexible `CSSObject`).

The change here looks pretty intentional but it might be worth an update to docs (the hover state suggested [here](https://react-data-table-component.netlify.app/?path=/docs/api-custom-conditional-formatting--page) and implemented in the story [here](https://react-data-table-component.netlify.app/?path=/docs/conditional-rows-rows--rows) does not work).

More restrictive typing might also help users avoid including styles which won't be applied.

PR #791, opened 2021-04-12
Added editable rows story

ahmetkuslular:
I added a row editing example. 😋

### DIFF
--- a/stories/DataTable/Basic/Editable.stories.js
+++ b/stories/DataTable/Basic/Editable.stories.js
@@ -1,4 +1,4 @@
-import React, { useRef, useState } from 'react';
+import React, { useCallback, useRef, useState } from 'react';
 import { storiesOf } from '@storybook/react';
 import data from '../constants/sampleMovieData';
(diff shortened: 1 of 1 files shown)


jbetancur:
Very nice 👍

stories/DataTable/Basic/Editable.stories.js
```diff
@@ -0,0 +1,148 @@
...
+            column={{ ...column, editing }}
+            col={col}
+            onChange={formOnChange}
+          />
+        );
+      },
+    };
+  });
+
+  const createColumns = () => {
```
jbetancur, line 115: I would recommend memorizing this function with `useCallback` to keep `DataTable` to minimize RDT re-rendering

jbetancur:
This is a really great example. Thank you for this. Just one comment for review

stories/DataTable/Basic/Editable.stories.js
ahmetkuslular: I made the fix. Thank you for your comments. 🥳

PR #793, opened 2021-04-15
Fixed double horizontal scrollbars

HoangNguyen17193:
Add minWidth to fixed  double horizontal scrollbars (large table with responsive, expandableRows, selectableRows options)

https://user-images.githubusercontent.com/16357791/114807906-7618e200-9dd1-11eb-8991-53293a00f7c6.mov


jbetancur:
Thank you for the fix!. There should definitely be a 48px width there. 

I cannot dupe this on my browser. Out of curiosity what browser are you using?

HoangNguyen17193:
@jbetancur , I'm using Chrome v90 on MacOS Big Sur

PR #804, opened 2021-05-13
Adding Column Filtering

AnotherCodeArtist:
We have added an option that allows to filter the data set by each column.
There's a new property in the column definition called `filterable`:
```javascript
const columns = [
	{
		name: 'Title',
		selector: row => row.title,
		sortable: true,
		filterable: true,
	},
       ...
]
```
If set to `true` a text input for every enabled column is rendered in the table header (Ok, design could probably be improved 😉). Works with server-side filtering as well!

Demonstration of this feature was added to the Kitchensink and to the Sorting stories.


AnotherCodeArtist:
Hi John!

All new code is covered by tests, nevertheless coverage is falling short by -0.03% 😢.
Anyway, you can check out the column filter demo here: https://anothercodeartist.github.io/react-data-table-component/?path=/story/general--kitchen-sink

MoosaSaadat:
@jbetancur is this feature expected to be added in coming releases?

davebrown1975:
As others have been asking, another request here for column filtering.

da7a90:
this should be added to the next release

PR #812, opened 2021-05-26
change hooks order execution when sorting

lalogrosz:
When you sort and you are in some page, the sort action reset the currentPage to 1 always.
Right now I don't want to add some logic to keep the currentPage, but I prefer to know first the sorting and then the changePage event, otherwise I don't know if the changePage is triggered by the pagination or by the sort.

Changing the useDidUpdateEffect order resolves this.


jbetancur:
Thanks for the PR. Do you see any cases where this might break other functionality? Do you have an example of what this is resolving for you?

lalogrosz:
No, I didn't see any other scenario.

PR #817, opened 2021-06-11
Fixed typo by replacing `Selected` to `onSelectedRowsChange` in README.md

patrice:
I think `onSelectedRowsChange` is the correct property to use to access the selected rows (rather than `Selected`).

Also, I removed the `* binary` in .gitattributes since it was preventing me from seeing text file diffs. I think the `* binary` line was intended for .gitignore. When used in .gitattributes, it means all files get checked out as type binary and git doesn't show diffs for binary files.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,2 @@
 # Set the default behavior, in case people don't have core.autocrlf set.
 * text=lf
-
-# Ignore any binary files
-* binary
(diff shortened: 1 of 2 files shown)


jbetancur:
NICE catch!

jbetancur:
note the build is failing due to a TS issue I am working on, so we can merge this in. Thanks for the PR!

PR #850, opened 2021-07-18
update README to remove XML declaration on ExpandableComponent

namtrhg:
remove XML declaration on ExpandableComponent as document metioned that it must be passed as ExpandableComponent not <ExpandableComponent />


namtrhg:
image

PR #867, opened 2021-08-14
Make column name optional

Dragate:
Make column name optional (as documented) so typescript projects don't demand this property


jbetancur:
Thank you!  LGTM!

Issue #872, opened 2021-08-17
Column sortFunction unclear in Readme


jgjr:
## Issue Check list
- [x] Agree to the [Code of Conduct](https://github.com/jbetancur/react-data-table-component/blob/master/CODE-OF-CONDUCT.md)
- [x] Read the README
- [x] You are using React 16.8.0+
- [x] You installed `styled-components`
- [x] Include relevant code or preferably a [code sandbox](https://codesandbox.io/embed/react-data-table-sandbox-ccyuu
)

## Describe the bug
The documentation for the sortFunction property of a column is unclear. Firstly, the link goes to #-custom-sort-function which does not work as the id of the section is #CustomSortFunction. Secondly, the custom sort function section (12) shows an example of a custom sortFunction for the table itself and not for an individual column. The sortFunction for an individual column accepts two arguments, rowA and rowB (as defined on line 6 of src/DataTable/types.ts), which is different from the table sortFunction. This might be obvious to many users, but it isn't explicitly stated in that section of the documentation and I had to discover this for myself.

## To Reproduce
1. Visit this GitHub repo
2. Read the Readme

## Expected behavior
I would have expected separate descriptions and examples for how the overall sortFunction and the column sortFunction work, as they are different.

jbetancur:
Thank you for calling this out. 

I will mention that `column[x]sortFunction` is only ever going to take 2 arguments because it is for a single `column` versus `sortFunction` which lets  you override sorting globally. `sortFunction` internally uses lodash `orderBy`, which should be fine 99% of the time.  On the other hand, `column[x]sortFunction` let's you override this global sorting for a specific column. A use case could be sorting semver strings (1.2.3, 2.3.4, etc) where typical sorting would not suffice. You would use `column[x]sortFunction` on the semver column to apply some special sort behavior to handle that.

 If you think this would be helpful for another user in the short term please consider creating a PR with changes to the README. Long term, I am in the process of "upgrading" the documentation to something better,

jbetancur:
Than you for the PR https://github.com/jbetancur/react-data-table-component/pull/873

merged

PR #873, opened 2021-08-17
Single column sortFunction documentation

jgjr:
I've updated the link in the columns section of the documentation and included a brief description and an example in the custom sort function section to explain how to define a custom sort function for a single column.


jbetancur:
NICE! Thank you!

closes #872

PR #956, opened 2021-11-01
Column filtering integrated

AnotherCodeArtist:
A new property called `filterable` makes a column filterable. Besides this, also server-side filtering is supported using property `filterServer` and callback `onFilter`.

Stories are also changed/extended to cover this new feature.


MDB22:
Hello! Wondering if/when this PR might be merged? RDT is amazing, this is the last big feature we need it to have! :)

Issue #954, opened 2021-10-28
Mistyping in Docs: Selectable/Pre Disabled Rows


afazrai:
react-data-table-component/stories/DataTable/selectableRows/preDisabled.mdx

Mistyping for <DataTable/> attribute:

Previously: 
`selectableRowSelected={rowDisabledCriteria}`
Correct one:
`selectableRowDisabled={rowDisabledCriteria}`

Some copy-pasting programmers might wonder why it won't pre disable the checkbox 👀

jbetancur:
Thanks for raising this. A PR is also welcome 😉

jbetancur:
Thank you! https://github.com/jbetancur/react-data-table-component/pull/960

PR #960, opened 2021-11-05
Fix mistyping for attribute #954

afazrai:
Correctly call the attribute to pre disable rows

Issue: Mistyping in Docs: Selectable/Pre Disabled Rows #954


jbetancur:
Nice catch. Than you!!!

PR #962, opened 2021-11-07
Allow styling header/title font-weight

FiveOFive:
Allow styling header/title font-weight

### DIFF
--- a/src/DataTable/TableHeader.tsx
+++ b/src/DataTable/TableHeader.tsx
@@ -21,7 +21,7 @@ const Title = styled.div`
 	flex: 1 0 auto;
 	color: ${({ theme }) => theme.header.fontColor};
 	font-size: ${({ theme }) => theme.header.fontSize};
-	font-weight: 400;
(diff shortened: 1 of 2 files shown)


jbetancur:
Thank you for the PR on this.  These couldalso inherit from `HeaderStyle` where `${({ theme }) => theme.header.style}` is being applied.

`Title` can just be:
```ts
const Title = styled.div`
	flex: 1 0 auto;
`;
```

and styles.ts

```ts
header: {
     style: {
	fontSize: '22px',
	fontWeight: 400, 
	color: theme.text.primary,
	backgroundColor: theme.background.default,
	minHeight: '56px',
	paddingLeft: '16px',
	paddingRight: '8px',
    },
},
```

Then you can just override font-weight on the `header` style.

Alternatively we could just create a new style named `headerTitle`


```ts
headerTitle: {
     style: {
	fontSize: '22px',
	fontWeight: 400, ,
    },
},
```
```ts

const Title = styled.div`
	flex: 1 0 auto;
       ${({ theme }) => theme.headerTitle.style};
`;
```

PR #970, opened 2021-11-25
Update patterns.stories.mdx

ziyadshaikh013:
Just a small change on line 53, we are expoting export default DataTableBase; instead it should be export default MyComponent;


jbetancur:
Thank you!

PR #1009, opened 2022-02-01
Added Selector and SortFunction to exported types

veddermatic:
Added these types to help those who use typescript to correctly type their sort functions.


jbetancur:
Nice thank you!

Issue #1007, opened 2022-01-31
[FEATURE]: Extensible props for each selectableRowsComponent instance


lautiab:
## Feature Check list

- [x] Agree to the [Code of Conduct](https://github.com/jbetancur/react-data-table-component/blob/master/CODE-OF-CONDUCT.md)
- [x] Read the README to ensure the feature is not already present
- [x] You read [Creating Issues, Features and Pull Requests](https://github.com/jbetancur/react-data-table-component/issues/387)
- [x] Considered the value versus complexity for all users of the library as well as library maintenance
- [x] Considered if this can be a storybook or documentation example

## Is your feature request related to a problem? Please describe

Right now on our a11y maintainability team we've been requested to customize aria attributes for each checkbox within rows, so that each checkbox it's labeled differently but with custom text (we need to use one data field, title in our case). So right now select-row-1 is not an aria-label we can have.

## Describe the solution you'd like

I think a modification of selectableRowsComponentProps to accept a function that receives the currently iterated row as param, and expects to output the props obj for that instance, would be a good way to solve this. That way you could have sth like..
`
selectableRowsComponentProps={commonPropsObj | (row) => { ...commonPropsObj, aria-label: row.title, otherSingularProp: row.whatEver }}
`
This solution is inspired in _cell_ prop API.

## Describe alternatives you've considered

The solution described above, or same approach but for selectableRowsComponent, but that might make selectableRowsComponentProps pointless since you would add your props to your JSX component directly.

## Additional context

Having an aria-label "select-row-1" might not be the greatest UX in many cases, so being more flexible regarding Checkbox props would fit better much more use cases.

PR #1011, opened 2022-02-14
Add possibility to opt-in for dynamic props for each checkbox instance.

lautiab:
Relates to [issue 1007](https://github.com/jbetancur/react-data-table-component/issues/1007)

### DIFF
--- a/src/DataTable/DataTable.tsx
+++ b/src/DataTable/DataTable.tsx
@@ -113,6 +113,7 @@ function DataTable<T>(props: TableProps<T>): JSX.Element {
 		customStyles = defaultProps.customStyles,
 		direction = defaultProps.direction,
 		onColumnOrderChange = defaultProps.onColumnOrderChange,
+		rowOnlyCheckboxProps = defaultProps.rowOnlyCheckboxProps
 	} = props;
 
 	const {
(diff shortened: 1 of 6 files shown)


lautiab:
Hey @jbetancur would you mind taking a look into this one? Thanks!

jbetancur:
@lautiab Not sure how I feel about this name `rowOnlyCheckboxProps` is should probably be named something like `selectableRowComponentProps`. This is a bit long, but it fits with the current naming convention. Note that in a version 8 of RDT I will be moving most of these props into an `options` prop. so this would in theory become `options.selectable.rowComponentProps`.

Also please add documentation to the storybook examples

jbetancur:
I wonder if another option could be to have `selectableRowsComponentProps` accept either object or callback format. This way we do not add a new property, but allow the you to opt into either format.

jbetancur:
For example,

`selectableRowsComponentProps={row => ({ 'aria-label': 'Select row ${row.title}', className: 'padding-1', etc... })}`

lautiab:
That was my first approach, but given how the prop is used for the heading checkbox too (select-all button), calling the function with row component as argument was resulting in undesired behavior for the header component props. Let me re-check those drawbacks so I can be more literal on my explanation, don't have the example right now, just recalling what I remember from the day I'd worked on it.

PR #807, opened 2021-05-18
Add onRowHovered event callback

TreyBastian:
This PR just adds a callback to allow actions to be taken when a row is hovered.

Our use-case was to prefetch data that that is normally requested when a user clicks on a row.

PR #1027, opened 2022-03-28
Add onRowMouseEnter and onRowMouseLeave event handlers

gsimko:
This PR picks up where the abandoned https://github.com/jbetancur/react-data-table-component/pull/807 was left and adds both the enter & leave events so that hovering can be properly detected.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -32,7 +32,7 @@
 		"editor.defaultFormatter": "esbenp.prettier-vscode"
 	},
 	"[typescriptreact]": {
-		"editor.defaultFormatter": "vscode.typescript-language-features"
+		"editor.defaultFormatter": "esbenp.prettier-vscode"
 	},
 	"[jsonc]": {
 		"editor.defaultFormatter": "vscode.json-language-features"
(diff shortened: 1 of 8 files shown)


gsimko:
@jbetancur can you please take a look at this PR?

.vscode/settings.json
```diff
@@ -32,7 +32,7 @@
 		"editor.defaultFormatter": "esbenp.prettier-vscode"
 	},
 	"[typescriptreact]": {
-		"editor.defaultFormatter": "esbenp.prettier-vscode"
+		"editor.defaultFormatter": "vscode.typescript-language-features"
```
jbetancur, line 35: Is this needed due to an update from prettier?

gsimko: No, it was an accidental change by vscode. I've reverted it.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-data-table-component",
-  "version": "7.4.7",
+  "version": "7.4.8",
```
jbetancur, line 3: This is a new feature so it should be 7.5.0

jbetancur: One last comment and this should be GTG

gsimko: done

PR #1037, opened 2022-04-14
Make paginationPage an optional, controllable property

bembleton:
- add paginationPage property to DataTable
- call handleChangePage when paginationPage changes
- maintain existing pagination controls and actions

I was looking for a way to reset the table back to the first page, and I was not happy with the way that `paginationResetDefaultPage` works. I prefer using a `value/onValueChanged` pattern so I would like to add a `paginationPage` property to control which page is selected when using either client-side or server-side pagination. It's expected that `paginationPage` and `onChangePage` are used together and kept in sync so that the pagination controls and actions still work:

```jsx
const [page, setPage] = useState(1);

<DataTable
	data={data}
	columns={columns}
	pagination
	paginationPage={page}
	onChangePage={setPage}
/>
```


GabrielLourenco:
Any release date?

MarcoDeJong:
Looking great!

garrylachman:
ERA for merge?

sadashivm:
Hello @jbetancur and team,
I'm unable to create new issue here hence I'm adding my doubts here only. please don't mind.
First of all, I want to say thank you so much for this react-data-table-component npm package. It has lot of features that you have added. It's really awesome.

My request is how to display header text as well as row data completely. Currently If header text / row cells data is greater than 30 to 40 character then you are showing ellipse (...). I know there is one column property like,
**grow : 5**

If I add above property in header column, then it's increasing the width of column. But is there any other way to display complete header text and row content without adding this property?.

afonsoogomes (approved):

mybess00:
Any update about this? There is any way to reset table back to the first page manually?

jbetancur:
Thank you for you patience on this feature. It's been added in v8.2.0 which ships today or early this week.

PR #1073, opened 2022-06-17
Fixed the table cell aria role to cell.

rukshan-rodrigo:
Each element with role="cell" MUST be nested in a container element with role="row". [Reference](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/row_role)
role=gridcell is only acceptable if we use role = grid in the parent. If we use role= table, we need to use the roll = cell in the table cell.


jbetancur:
Nice catch! Thank you!

PR #1076, opened 2022-06-21
Adding className prop to ResponsiveWrapper

ZakariaBenali:
The className prop is defined in TableProps but it's not used in the table at all, which makes editing the style of the table via CSS requires wrapping the table with another wrapper. this is fixed by passing down the className that already exists on TableProps to ResponsiveWrapper.


bjermeland:
Can this PR be completed?

jbetancur:
Yes and Thank you!

jbetancur:
These changes will be available in a patch release by this evening.

PR #1105, opened 2022-08-26
Fix className for non-sortable columns

pallanson:
Even when a column is not sortable, the className "rdt_TableCol_Sortable" still exists.


pallanson:
@jbetancur Any chance you could take a look at this? Seems like a bug but I'm not sure why all the checks failed.

pallanson:
@jbetancur Just want to check in again and see if we can get this fixed.

MarcoDeJong:
Need help on this?

PR #1122, opened 2022-09-30
Use semantic html elements for accessibility

nikolas:
We received an accessibility audit for one of our projects using react-data-table-component. Although this library uses the `role` attribute appropriately, there's no reason for it to be using `div` instead of `table`. This would really help screen readers know what to do with these tables.

Does a change like this make sense?


nikolas:
Thanks for your feedback. I agree - this is a bigger change than this PR. To start, I'd have to fix all the test failures.

To resolve our specific issue, because this library is already using the role="table" attributes properly, we are considering it accessible for our purposes. Screenreaders should be able to handle these divs with ARIA roles just fine. Though, as I've learned, even Mozilla recommends using `<table>` over ARIA roles whenever possible: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/table_role#required_javascript_features

PR #1130, opened 2022-11-04
feat: add expandable - GroupBy Data

EliseiNicolae:


stories/DataTable/expandable/groupBy.stories.js
```diff
@@ -0,0 +1,106 @@
+import React from 'react';
+import * as _ from 'lodash-es';
+
+import DataTable from '../../../src/DataTable/DataTable';
+import Row from '../../../src/DataTable/TableRow';
+import tableDataItems from '../../constants/sampleMovieData';
+
+const ExpandedComponent = ({ data, rowProps }) => {
```
iosifnicolae2, line 8: I would move this inside the library, so the user will have to pass GroupByKey and this component is used as default  `expandableRowsComponent` (if the user wants to override it, then he/she will pass his own `expandableRowsComponent` property).

stories/DataTable/expandable/groupBy.stories.js
```diff
@@ -0,0 +1,106 @@
...
+			expandableRowsComponent={ExpandedComponent}
+		/>
+	);
+};
+const Template = args => <GroupByStory {...args} />;
+
+export const GroupBy = Template.bind({});
+
+GroupBy.args = {
+	expandableRows: true,
```
iosifnicolae2, line 92: also, this property should be set automatically to `true` inside the library when passing `groupKey`

stories/DataTable/expandable/groupBy.stories.js
```diff
@@ -0,0 +1,106 @@
...
+	pagination: true,
+	groupKey: row => `Year: ${row.year} - Runtime: ${row.runtime}`,
```
iosifnicolae2, line 94: I would rename this to `groupBy`

stories/DataTable/expandable/groupBy.stories.js
```diff
@@ -0,0 +1,106 @@
...
+	groupValues: data => {
```
iosifnicolae2, line 95: I would rename this to `groupValue` and I would rename `data` to `groupData`

stories/DataTable/expandable/groupBy.stories.js
```diff
@@ -0,0 +1,106 @@
...
+		name: 'Count',
+		selector: row => row.count,
+		sortable: true,
+		wrap: true,
+	},
+];
+
+const GroupByStory = props => {
+	const dataByGroupKey = _.groupBy(tableDataItems, row => props.groupKey(row));
+	const groupedData = _.map(dataByGroupKey, (group, key) => {
```
iosifnicolae2, line 69: I would move this logic inside the library when the user uses `groupBy` property.

stories/DataTable/expandable/groupBy.stories.js
```diff
@@ -0,0 +1,106 @@
...
+				name: 'Runtime',
+				selector: row => row.runtime,
+				sortable: true,
+				wrap: true,
+			},
+		],
+	};
+	return _.map(data.groupData, item => {
+		return (
+			<div key={item.title}>
```
iosifnicolae2, line 45: this `div` is not required, we can pass the `key` attribute to `Row` component
PS. make sure that the key value is unique (`item.title` might have duplicate values)
Also, `item.title` might be undefined.

A good idea might be to check out how datatable generate the `key` value and use the same mechanism for generating keys for group rows.

iosifnicolae2:
### Component Header
```jsx
<DataTable
     title="Movie List Grouped"
     data={groupedData}
     columns={columns}
     groupKey={row => `Year: ${row.year} - Runtime: ${row.runtime}`}
     groupValues={data => {
      return {
         count: data.length,
     };
/>
```

PR #1140, opened 2023-01-19
Adding Column Header Filter

AnotherCodeArtist:
Like several times before ...

That's probably the only missing key feature, but if you are fine with it that's ok with me as well.


sadashivm:
Hello @jbetancur ,
Currently there is each column filter option is there, but do we have global table filterable instead of each column?
I have checked in Kitchen sink examples, I'm not sure what is the use of below search box.

image

It is not filtering anything from the table.

I need to search any column data from single serachbox input . How to achieve ?

Is it possible?

sadashivm:
Hello all and @jbetancur ,
When we select records from the table, selected rows are not pushing into array sequentially.
If I select below rows like,
15th row, 8th row, 20th row and 4th row then I'm expecting final array should be,

**[
  {15th row object},
  {8th row object},
  {2oth row object},
  {4th row object}
]**

But data table return different array in different order like,

**[
  {4th row object},
  {8th row object},
  {15th row object},
  {20th row object}
]**

DO you any solution?

sadashivm:
@jbetancur , @AnotherCodeArtist  any update on above issues?
DO you have any solution on header column filter and second requirement is,
push selected rows from the table in sequential order I mean user select 10th row, 5th row and 16th row then selectedRows array must return array with [10th row, 5th row, 16th row] in this order

frankdilu:
any update? this is useful

jbetancur:
The PR is appreciated, but we went in a different architectural direction in v8 which now implements filtering in headers

PR #1142, opened 2023-02-02
Allow custom row container as optional

ashotmark:
## Why

## What

## Example

```
```


ashotmark:
@nickgirardo  Thanks for comment
I created storybook for this usages, so that you can see how to use on your component.
in my case, I installed this package from  current `github` repo because we can not wait until this PR gets approved and merged 👍 
here is storybook: `stories/DataTable/RowLink.stories.js`

SupraSmooth:
I would love to have that feature! Please run `eslint --fix \"**/*.*\"` and `prettier --write \"**/*.*\"` commands in your project directory to fix the code style errors. Run `npm run lint` before your push to verify your code.

ashotmark:
Thanks @SupraSmooth , I added it towork list, will fix and push today.

ashotmark:
fixed eslint and prettier issue. now all checks are passed :100:

paulStoia:
Great job with this feature! When it will be merged and released? It is very useful and critical in some cases because it allows to control how a row is rendered, hence you can add much more operations in the table. We really need this!

samuelhalim1:
any updates?

Niwilai:
@jbetancur any chance we could get this in? It would be useful for us. Ty

PR #1144, opened 2023-04-30
Support for right click on rows

franverona:
This PR contains:

- [ ] bugfix
- [x] feature
- [ ] refactor
- [ ] documentation
- [ ] other

### Description

Add support for right click on table rows for allowing row context menu logic.

```js
function MyTable({ columns, data }) {
	const handleRowRightClicked = row => {
		console.log(`${row.name} was right clicked!`);
	};

	return (
		<DataTable
			columns={columns}
			data={data}
			onRowRightClicked={handleRowRightClicked}
		/>
	);
}
```


sadashivm:
Hi @franverona ,
I'm sadashiv I'm unable to create a new issue here and for me create "new issue" button is not appearing anywhere.

Here I have added comment for two issues,

[https://github.com/jbetancur/react-data-table-component/pull/1140](url)

Do you have solution for above requirements?

1. I need input search filter for each column / global search filter (It can search any column data)
2. Selected rows not pushing in sequential order.

Kindly let me know, if you need further details

PR #1145, opened 2023-06-09
feat: exposing col sort state as CSS class

david-ampsec:
Current implementation does not expose column sort state. The styling is hidden behind styled components. This PR adds an `active_sort` class to columns when using a custom `sortIcon`.


land0r:
@jbetancur hey, could you approve that pull request?

land0r (approved):
Seems good for me, wondering why this pull request is still "Open"

PR #1148, opened 2023-07-09
Added Chekbox position feature to dataTable.

razma06:
Added a feature. One prop "selecrtableRowsPosition" which is enum of "left" and "right".
You can use this prop to change the position of checkbox column in a table.


jbetancur:
closing stale PR

PR #1151, opened 2023-07-10
Feat: Pagination option last and first page

dipras:
I need to hide first page and last page pagination button. Because the current code not provide options to hide it and can only be hidden if all component is replaced. So i modif the code and open this pull request, so i hope the people need this feature will happy :)


jbetancur:
This needs tests and a storybook example

PR #1155, opened 2023-08-21
Fix styled component errors 2

jessewriter:
building off off [HendrikPetertje](https://github.com/HendrikPetertje) PR i also removed the final styled component errors for 'right' and 'sortActive'

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -38,4 +38,9 @@
 		"editor.defaultFormatter": "vscode.json-language-features"
 	},
 	"prettier.useTabs": true,
+	"workbench.colorCustomizations": {
+		"activityBar.background": "#320B88",
+		"titleBar.activeBackground": "#460FBE",
+		"titleBar.activeForeground": "#FCFAFF"
+	},
 }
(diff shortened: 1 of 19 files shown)


jessewriter:
not sure why there are netlify errors as this work just fine in my project and all the styled warnings are gone from my console.  we have many different tables and no errors :/

jessewriter:
ok i see that my $right fix did not work and those columns are still left oriented.  error is gone but intended effect is wrong will investigate tmro

jessewriter:
> heya @jessewriter, noticing the error you are trying to fix too now :) Do you want me to help you out so we can get this pull over the line?

yes sir, we are running this fork successfully but i never did update the unit tests.  right and sortActive both cause issues and we fixed.  can't remember if there were any more but 0 errors on our side.  only unit tests remain

amirfaisalz:
any update on this?
the issue still persist

SamuelDowse:
The issue is with styled-components version 6.x and above, the issue does not exist on styled-components version 5.x

This library needs updating to support styled-components version 6.x

allan-vera:
Are there any suppression suggestions besides installing the correct version of style-components (for projects that dont use SC)

Issue #1156, opened 2023-08-24
react-dom.development.js:86 Warning: React does not recognize the `sortActive` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `sortactive` instead. If you accidentally passed it from a parent component, remove it from the DOM element


Arjunv17:
When I am using latest version that I am facing this issue. Please fix it and let me know any suggestion how can I fix this.

Thanks

![image](https://github.com/jbetancur/react-data-table-component/assets/100670181/15dad7c1-d376-44b7-a49d-d2ae523498a2)
![image](https://github.com/jbetancur/react-data-table-component/assets/100670181/ad69ad67-9f12-4873-8bba-d6931ee9baac)

Here Is my code :
    const columns = [
        {
            name: 'Sr No.',
            selector: (row, index) => index + 1,
        },
        {
            name: 'From',
            selector: (row) => row.from,
        },
        {
            name: 'To',
            selector: (row) => row.to,
        },
        {
            name: 'Date',
            selector: (row) => row.leave_date,
        },
        {
            name: 'Leave Type',
            selector: (row) => row.type,
        },
        {
            name: 'Message',
            selector: (row) => row.message,
        },
        {
            name: 'Status',
            selector: (row) => row.status,
        },

    ];

             <DataTable
                      columns={columns}
                      data={data} // data on api response 
                  />

amirfaisalz:
any update on this?
the issue still persist

Arjunv17:
YES, this issue is still occurring.

react-dom.development.js:86 Warning: React does not recognize the
`sortActive` prop on a DOM element. If you intentionally want it to appear
in the DOM as a custom attribute, spell it as lowercase `sortactive`
instead. If you accidentally passed it from a parent component, remove it
from the DOM element.

SamuelDowse:
I've hit this issue as well, looks like it's caused by using styled-components library version 6.x.
If you install styled-components@5.3.11 it'll get rid of these errors.

Styled-components switched from camelCase to lowercase for their props, and react-data-table-component library will need updating to support styled-components@6.x

Arjunv17:
Fixed!
Thanks for the help!!

SamuelDowse:
Not sure why my comment deleted, but to summarise for anyone else looking at this thread in the future until this bug is resolved. 

[...]

PR #1192, opened 2024-01-04
Fix sortActive like an unknown prop using styledComponent transient props

JacksonSamuel42:
fix "sortActive" is being sent through to the DOM, which will likely trigger a React console error. If you would like automatic filtering of unknown props.

fix this problem using transient props https://styled-components.com/docs/api#transient-props


JacksonSamuel42:
the pr of problem is #1156

jbetancur:
Thank you for this PR, however, a major update was required as part of a refactor I have been working on and this was fixed in https://github.com/jbetancur/react-data-table-component/pull/1195

PR #1206, opened 2024-01-20
Fix width of table columns and fix some mistakes in Storybook (Netlify)

ADTC:
Some of the tables are really wide due to non-wrapping code blocks. The columns were also not distributed well due to large code blocks in Type and Default columns. There were also several mistakes in the content.

**This pull request:**

* Fixes the width of table columns by allowing wrap on code blocks.
* Unsets the line-height of code blocks, to make them look nicer.
* Fix some mistakes in code like missing brackets, missing lambda arrows, missing code blocks, lost cell, etc.
* Move the longer Type and Default column values into the Description. This better distributes the table columns so the description has more space.
  * Where moved, the original cell has _"See description"_ in it now for guidance.

**Preview:**

* [Columns](https://deploy-preview-1206--react-data-table-component.netlify.app/?path=/docs/api-columns--docs)
* [Properties](https://deploy-preview-1206--react-data-table-component.netlify.app/?path=/docs/api-props--docs) _(major changes on this page - [compare live site](https://react-data-table-component.netlify.app/?path=/docs/api-props--docs))_
* [Conditional Styles](https://deploy-preview-1206--react-data-table-component.netlify.app/?path=/docs/api-custom-conditional-formatting--docs)


ADTC:
@jbetancur just let me know if I missed something.

.storybook/base.css
```diff
@@ -7,3 +7,8 @@ body {
 table {
     width: 100%;
 }
+
+code.css-1kwwth4 {
```
jbetancur, line 11: You would want to set this on the styled component rather than the generated css class name.

ADTC: That's a good question. So far in my checks it's always the same name that's generated. It never changes.

I can just have it as a tag only since the other declarations are doing the same in this CSS file. As in `code { ... }`. What do you think?

ADTC: Hi @jbetancur I will change it to just `code` as said. Hope to get this merged because the tables are too wide right now. I'll do this soon today.

ADTC:
@jbetancur I've removed the generated class name now. Note: The `!important` keyword is necessary because the generated class name that applies overriding styles has a higher specificity than the element tag selector, so we need to force-override it.

ADTC:
Thank you! I appreciate it.

PR #1218, opened 2024-03-21
Allow custom row container to wrap table row.

ashotmark:
## Note:
This is re-PR for the [Allow custom row container as optional](https://github.com/jbetancur/react-data-table-component/pull/1142)
I send PR again because former PR was closed due to inactivity, but I think this is still worth to consider merge because this helps some projects and never impact to existing features.
I rebased changes from upstream master, so this should be up to date.

## Why

## What

## Example

```
```


ashotmark:
Close due to wrong PR

PR #1219, opened 2024-03-21
Allow custom row container to wrap table row with prefer tag or component

ashotmark:
## Note:
This is re-PR for the [Allow custom row container as optional](https://github.com/jbetancur/react-data-table-component/pull/1142)
I send PR again because former PR was closed due to inactivity, but I think this is still worth to consider merge because this helps some projects and never impact to existing features.
I rebased changes from upstream master, so this should be up to date.

## Why

## What

## Example

```
```


robotic-forest:
Important feature for me as well, big ask from my users. Any blockers here? Happy to help test @jbetancur

robotic-forest:
@ashotmark this feature works great. I'm using your forked version for now (hope you don't mind lol) and have had no problems. For anyone else seeing this, I'm just using it in next.js like this without any hiccups:
```
renderRow={(row: any, rowContent: any) => <Link key={row.id} href={`/appointments/${row.id}`}>{rowContent}</Link>}
```

jbetancur:
Hey, thanks for the PR and sorry for the slow response!

I can see the use case but I'm not sure this adds enough value to justify the new API surface. Most row navigation scenarios are covered by onRowClicked — and wrapping rows in an <a> tag has real HTML validity and layout concerns that would be on the user to manage.

Can you give me some examples of what this looks like visually?

jbetancur:
closing due to inactivity

PR #1224, opened 2024-04-08
Fix Select component styling and remove duplicate Dropdown Icon

memmen:
This pull request fixes the styling of the Select component by removing the cursor and removes the unused DropdownIcon component.


jbetancur:
This is working as designed as the DropDownIcon is totally used in the Select control.

Issue #649, opened 2020-08-12
[BUG] When using SSR I get a dom mismatch


izaacdb:
## Describe the bug
When using NextJS I see 
`Warning: Expected server HTML to contain a matching <span> in <nav>.
    in span (created by Styled(styled.span))
    in Styled(styled.span) (created by Pagination).`
This warning only shows up when I use pagination.

## To Reproduce
1. npx create-next-app --example with-typescript-styled-components with-typescript-app 
2. yarn add react-data-table-component
3. copy the example data table, add 11 records
4. enable pagination
5. yarn dev
6. see error in console, compare view source to rendered code

## Expected behavior
`<span class="sc-fznxKY sc-fznMAR jMFmnk">1-10 of 537</span>` should be rendered on server and client

## Versions (please complete the following information)
 - next@9.5.1
 - styled-components@5.1.1
 - react-table@7.5.0

erikt9:
I've been unable to get SSR working with style-components@5, you may want to try version 4 and see if you have any luck with that.  Also make sure to be using the `babel-plugin-styled-components` with your build if appropriate.

izaacdb:
v5 seems to work fine for me other than this issue. I am using that plugin and some Babel configuration.

jbetancur:
TBH I have next to no experience using next.js. Is this just a warning or do things actually break Pagination?  I'm really curious if there is some side effect with styled-components composition and nextjs. For example, the span that has the rows count `Range` is derived from a command span.

https://github.com/jbetancur/react-data-table-component/blob/5fa8a9bb93d7c2bcc861ab256130c84cbb4cf101/src/DataTable/Pagination.js#L50-L57

abdulsamadayoade:
Is there any fix for this @jbetancur @izaacdb . I've been facing this same issue for sometime now

kayotimoteo:
same problem here

Issue #1103, opened 2022-08-22
Using in NextJs project，but got "Expected server HTML to contain a matching <span> in <nav>. "


yourmoonlight:
## Describe the bug

I have no idea why this error came out. 

Warning: Expected server HTML to contain a matching <span> in <nav>.
    at span
    at O (webpack-internal:///./node_modules/styled-components/dist/styled-components.browser.esm.js:31:19811)
    at nav
    at O (webpack-internal:///./node_modules/styled-components/dist/styled-components.browser.esm.js:31:19811)
    at eval (webpack-internal:///./node_modules/react-data-table-component/dist/index.cjs.js:396:40)
    at div
    at Fe (webpack-internal:///./node_modules/styled-components/dist/styled-components.browser.esm.js:31:17360)
    at eval (webpack-internal:///./node_modules/react-data-table-component/dist/index.cjs.js:396:6011)
    at div
    at PeerTable (webpack-internal:///./pages/peers/index.js:35:36)
    at main
    at Layout (webpack-internal:///./components/layout/layout.js:19:33)
    at MyApp (webpack-internal:///./pages/_app.js:42:27)
    at ErrorBoundary (webpack-internal:///./node_modules/next/dist/compiled/@next/react-dev-overlay/client.js:8:20746)
    at ReactDevOverlay (webpack-internal:///./node_modules/next/dist/compiled/@next/react-dev-overlay/client.js:8:23395)
    at Container (webpack-internal:///./node_modules/next/dist/client/index.js:323:9)
    at AppContainer (webpack-internal:///./node_modules/next/dist/client/index.js:825:26)
    at Root (webpack-internal:///./node_modules/next/dist/client/index.js:949:27)

abdulsamadayoade:
I'm facing the same issue too

Kerotobil:
I had a some issue and solved like this
Match the Pagination to router.isReady

```
 const router = useRouter();



   <DataTable
            title="Countries"
            columns={columns}
            data={filteredData}
            defaultSortFieldId={1}
            selectableRows
            pagination={router.isReady}
            paginationServer={false}
            paginationIconFirstPage
            onSelectedRowsChange={(e) => handleSelected(e.selectedRows)}
          />
```

PR #1227, opened 2024-04-15
Add `paginationComponentOptions.minWindowWidth` (defaults to `SMALL == 599`)

ryan-williams:
## [Conditional rendering in `Pagination`](https://github.com/jbetancur/react-data-table-component/blob/v7.6.2/src/DataTable/Pagination.tsx#L98) causes SSR hydration errors in Next.js apps
- Repro repo [runsascoded/react-data-table-hydration-bug](https://github.com/runsascoded/react-data-table-hydration-bug)
- [Live demo](https://react-data-table-hydration-bug.runsascoded.com/) (refresh with different windows widths, dev console shows error iff  `window.innerWidth ≥ 599`).
- [Local dev mode](https://github.com/runsascoded/react-data-table-hydration-bug?tab=readme-ov-file#local-repro) gives more details:
  [error](https://github.com/runsascoded/react-data-table-hydration-bug/blob/main/public/error.gif)
- The same error is described by #649 and [this StackOverflow post](https://stackoverflow.com/q/75068071).

## Workaround / Fix
This PR adds the ability to pass `paginationComponentOptions={{ minWindowWidth: null }}` to `DataTable`, to always render the "rows per page" and "page range" parts of the pagination footer, which ensures server and client DOMs will match.

### DIFF
--- a/src/DataTable/Pagination.tsx
+++ b/src/DataTable/Pagination.tsx
@@ -95,7 +95,9 @@ function Pagination({
 }: PaginationProps): JSX.Element {
 	const windowSize = useWindowSize();
 	const isRTL = useRTL(direction);
-	const shouldShow = windowSize.width && windowSize.width > SMALL;
+	const { minWindowWidth = SMALL } = paginationComponentOptions;
+	const windowWidth = windowSize.width;
+	const shouldShow = !minWindowWidth || (windowWidth && windowWidth > minWindowWidth);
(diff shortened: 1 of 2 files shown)


ryan-williams:
Just noticed this addresses #1103

jbetancur:
#1103  should have addressed this - if not please re open an issue/PR

Issue #1208, opened 2024-01-24
Add Footer Row Support


manishagupta016:
I propose adding support for a customizable footer row at the bottom of tables. This feature would enhance data presentation, allowing users to display summary information or totals conveniently in the library's tables

jbetancur:
I know this has been open for quite a while so thank you 8.2.0 coming soon will ship with a footer feature

PR #1235, opened 2024-05-30
Add support for a footer row

jcognard-actinvision:
## Why
For one of my projects I need to add a footer row to my table and make it configurable by users to display sums, averages and more.
There’s already someone asking for a footer row support in #1208 

## What
This PR aims to enable support for a footer row by accepting a `showFooter` boolean attribute on `DataTable` component.
The content of the footer cells are defined with the `footerContent` function on the columns properties, which receives all visible rows as an argument, making possible to calculate sums, for example.

## Example
```javascript
import tableDataItems from '../constants/sampleDesserts';
import DataTable from '../../src/index';

const columns = [
	{
		name: 'Name',
		selector: row => row.name,
		sortable: true,
		fixed: true,
	},
	{
		name: 'Type',
		selector: row => row.type,
		sortable: true,
		fixed: true,
	},
	{
		name: 'Calories (g)',
		selector: row => row.calories,
		sortable: true,
		right: true,
		footerContent: rows => {
			const total = rows.reduce((acc, row) => acc + row.calories, 0);
			return Math.round(total);
		},
	},
];

export const Footer = () => {
	return <DataTable title="Movie List" columns={columns} data={tableDataItems} showFooter />;
};
```

I hope the quality of code matches your standards as it’s my first contribution.

### DIFF
--- a/src/DataTable/DataTable.tsx
+++ b/src/DataTable/DataTable.tsx
@@ -1,37 +1,41 @@
 import * as React from 'react';
 import { ThemeProvider } from 'styled-components';
-import { tableReducer } from './tableReducer';
-import Table from './Table';
-import Head from './TableHead';
-import HeadRow from './TableHeadRow';
-import Row from './TableRow';
-import Column from './TableCol';
-import ColumnCheckbox from './TableColCheckbox';
-import Header from './TableHeader';
-import Subheader from './TableSubheader';
-import Body from './TableBody';
-import ResponsiveWrapper from './ResponsiveWrapper';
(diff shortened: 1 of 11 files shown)


jbetancur:
@jcognard-actinvision I've been a "little slow" to respond but I will take a look at this for consideration and review

jbetancur:
Hey, @jcognard-actinvision  thanks for the PR and sorry for the slow response!

We've moved in a different architectural direction for v8 and this feature was implemented as part of the update. Take a look and see if the new footer works for you.  Thanks again for the contribution.

Issue #1203, opened 2024-01-15
Boolean column property button triggers console warning


raffaele-clevermind:
## Describe the bug
After updating to version 7.6.2, which should support styled-components ^6.0.0, when using the boolean collun property button it generates a warning as reported below

Received `true` for a non-boolean attribute `button`.
If you want to write it to the DOM, pass a string instead: button="true" or button={value.toString()}.

## To Reproduce
1. Intall react-data-table-component ^7.6.2
2. Install styled-components ^6.1.8
3. Use a table component with a column with the proprerty "button" set to true
4. in the browser console check for the warning as described iin the description of the issue

## Expected behavior
No warnings in the console

randydijkstra:
Not all CellProps properties are updated to transient props yet. As far as I can find most other components are fine.

ezequielramos:
Is there some fix or workaround to make this warning/error stop showing? It doesn't really seems that there's a problem happening

Hazbob:
was a workaround found for this? our tests are getting serious spam from it and i dont want to suppress the console logs :)

Bohemme:
I think it's similar to compact attribute, which should be boolean but throws a warning. I've used  compact: 1 instead and works just fine. :)

jbetancur:
Closing as this issue has been open since 2023–2024 with no resolution. v8 is now available and includes significant architectural changes that may have already resolved this — please check the [v8 docs](https://reactdatatable.com) and [changelog](https://reactdatatable.com/docs/changelog) first. If the issue persists on v8, please open a new issue with a reproduction against the current version.

PR #1239, opened 2024-07-02
remove Warning on the table

ferchuhill:
### Why
---
I want to remove this warning: https://github.com/jbetancur/react-data-table-component/issues/1203

```
VM26934:1 Warning: Received 'true' for a non-boolean attribute 'center'. 
If you want to write it to the DOM, pass a string instead: center="true" or center={value.toString()}.
```

### What
---
Make some change on the types and where is calling so is passing the like a string.

### Example
---

```
const columns = [
	{
		name: 'Name',
		selector: row => row.name,
		sortable: true,
		center: true,
	{
];

```

### DIFF
--- a/src/DataTable/TableCell.tsx
+++ b/src/DataTable/TableCell.tsx
@@ -67,9 +67,9 @@ function Cell<T>({
 			$cellStyle={column.style}
 			$renderAsCell={!!column.cell}
 			$allowOverflow={column.allowOverflow}
-			button={column.button}
-			center={column.center}
-			compact={column.compact}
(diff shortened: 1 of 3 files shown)


jbetancur:
Thank you for the PR, however, I would rather not add code for this as it is preferable IMO to have the user correct this rather than RDT accounting for every case of type coercion

dief:
The typescript definition only allows it to be boolean, not string. Can the type definition be changed for this?

Hazbob:
was there any resolution to this? as the columns types expect a boolean, but that seems to be causing the issue

Issue #1272, opened 2025-02-05
Add Accessibility Support for Table Titles (aria-label)


deiberchacon:
**Overview:**
Currently, `react-data-table-component` provides a `title` prop, but screen readers do not inherently have context about the meaning or purpose of the table itself. I propose adding an option to use `aria-label` to provide this context for screen reader users.

**Proposed Solution:**
- **Option 1:** Use the existing `title` prop to set an `aria-label` on the table element.
  ```jsx
  <DataTable title="Quarterly Report" data={data} columns={columns}  />
  ```
  Should render: 
  ```html
  <div class="rdt_Table" role="table" aria-label="Quarterly Report">
      ...
  </div>
  ```
  Screen reader will read "Quarterly Report table"

- **Option 2:** Add a new `aria-label` prop to directly define the table’s label for screen readers.
  ```jsx
  <DataTable title="Quarterly Report" data={data} columns={columns} aria-label="Quarterly Financial Report" />
  ```
  Should render: 
  ```html
  <div class="rdt_Table" role="table" aria-label="Quarterly Financial Report">
      ...
  </div>
  ```
  Screen reader will read "Quarterly Financial Report table"

**Pros:**
- Provides clear context about the table’s meaning for screen reader users.
- No breaking changes for existing users.

**Cons:**
- Minor design change, but requires good documentation to ensure clarity.

jbetancur:
closed by #1273

PR #1273, opened 2025-02-11
Add support for table aria-label (accessibility)

deiberchacon:
### Why

Screen readers and accessibility tools do not inherently have context about the meaning or purpose of the table itself.

I opened a feature request for this https://github.com/jbetancur/react-data-table-component/issues/1272 and had some time to work on it.

### What

This PR adds an optional `ariaLabel` string property to the table component.  This property sets the `aria-label` attribute, providing context for screen readers.  If `ariaLabel` is not provided, the `aria-label` attribute will not be added.  

This change has no visual impact, just useful for accessibility.


jbetancur:
Thank you for the feature/pr! LGTM!

PR #1289, opened 2025-08-05
Feature/accessibility for selactable row

sudhansurana:
Added roles to checkbox container


jbetancur:
Thanks for the contribution and apologies for the long wait.

v8 has since shipped with a different architecture, and this functionality is now available out of the box. Closing this PR as it no longer applies to current master.

If you're still using v7 you're welcome to keep this on your fork, but v7 is no longer maintained. I'd encourage upgrading to v8 — see the [docs](https://reactdatatable.com/) for the migration.

Thanks again for the effort!

Issue #1159, opened 2023-08-29
[FEATURE]: [SHORT DESCRIPTION]


mikematos84:
## Feature Check list

- [x] Agree to the [Code of Conduct](https://github.com/jbetancur/react-data-table-component/blob/master/CODE-OF-CONDUCT.md)
- [x] Read the README to ensure the feature is not already present
- [x] You read [Creating Issues, Features and Pull Requests](https://github.com/jbetancur/react-data-table-component/issues/387)
- [x] Considered the value versus complexity for all users of the library as well as library maintenance
- [x] Considered if this can be a storybook or documentation example

## Is your feature request related to a problem? Please describe

Allow for disabling of tabbing on empty header cells so they do not get read by screen readers

## Describe the solution you'd like

Allow the passing of `tabindex` to be passed through to `TableCol` cell so the user gains greater flexibility, or simple set `tabindex="-1"` if the content is empty

BackupMiles:
I agree with this issue, I would also include a default behaviour: if a header cell has set "sortable" to false, a `tabindex` should not be applied imho

kenirwin:
The accessibility angle @BackupMiles mentioned is an important one. The "main" issue I'd put in the "it might be nice" column, but for accessibility `tabindex="-1"` when not sortable should be a priority.

PR #1291, opened 2025-09-19
use tabindex="-1" on non-sortable column headers

kenirwin:
The proposed fix sets the `tabindex="-1"` (unfocusable) wherever the "disabled" property is set on the column-sorting functionality.

This is an accessibility fix; currently, all column headers on DataTable have `tabindex="0"`, making them focusable. That's appropriate for interactive components like sortable column headers, but where sorting is disabled, it adds unnecessary keystrokes and disrupts the normal flow of selection on the page (violating [WCAG 2.4.3](https://www.w3.org/WAI/WCAG21/Understanding/focus-order.html))

I've updated the test snapshots for DataTable to reflect these change -- there are 77 cases where what used to have a `tabindex="0"` is now `tabindex="-1"`.

Note: This fixes an issue raised in the comments to #1159 but is not a fix for the overall issue the original poster discussed. (They wanted more direct user control over the tabindex for each column; this PR just obviates one potential use-case for that enhancement, and improves on it by enforcing WCAG compliance rather than relying on the component's installer to do so.)

### DIFF
--- a/src/DataTable/TableCol.tsx
+++ b/src/DataTable/TableCol.tsx
@@ -177,6 +177,7 @@ function TableCol<T>({
 
 	const sortActive = !!(column.sortable && equalizeId(selectedColumn.id, column.id));
 	const disableSort = !column.sortable || disabled;
+	const tabIndex = disableSort ? -1 : 0;
 	const nativeSortIconLeft = column.sortable && !sortIcon && !column.right;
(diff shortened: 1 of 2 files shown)


kenirwin:
@jbetancur - thanks for merging this PR. I wonder if you'd be willing to issue a patch-release that includes this accessibility update. I tried building the repo locally as a workaround under your next release is ready, but I wasn't able to get it to build. Many thanks!

PR #1316, opened 2026-05-17
feat: add controlled paginationPage prop and update API documentation

jbetancur:


jbetancur:
implemented in 8.2.0!

PR #1375, opened 2026-08-16
fix(selection): use the controlled selectedRows prop as the delta base

### LINKED ISSUE #1374: Controlled selectedRows is dropped when toggling another row in-table
When selection is controlled via `selectedRows`, toggling a row inside the table computes the change from the component's **internal** selection rather than the controlled prop. Because internal state is never reconciled with the prop, any selection set from outside the table (an initial value, or an update from another component) is lost on the next in-table toggle, and `onSelectedRowsChange` emits the wrong set.

## Reproduction

```tsx
const [selected, setSelected] = useState([]);
useEffect(() => setSelected([data[0]]), []); // select row 1 externally

<DataTable
  keyField="id"
  data={data}
  columns={columns}
  selectableRows
  selectedRows={selected}
  onSelectedRowsChange={({ selectedRows }) => setSelected(selectedRows)}
/>
```

Tick row 2's checkbox.

- **Expected:** `[row1, row2]`
- **Actual:** `[row2]` — row 1 is dropped, because the delta is computed against the stale internal `[]`.

## Cause

`selectedRows` is applied only as a display projection on the reducer's output; the reducer still computes `SELECT_SINGLE_ROW`/`SELECT_RANGE`/`SELECT_ALL_ROWS` from its own `state.selectedRows`, which is never reconciled with the prop.

## Version

8.8.0

noahisdai:
## Summary

Controlled selection deltas are computed from internal reducer state, which is never reconciled with the `selectedRows` prop — so an externally-set selection is dropped on the next in-table toggle and `onSelectedRowsChange` emits the wrong set. This wraps the reducer so that, when controlled, it computes against the prop instead of stale internal state. Uncontrolled behavior is unchanged.

Fixes #1374

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation / example update

## Checklist

- [x] I read CONTRIBUTING.md
- [x] I agree to the Code of Conduct
- [x] Tests pass (`npm test`)
- [x] I have updated docs or examples if needed (no doc change required — behavior of the existing `selectedRows` prop)

## How to test

1. Render a `DataTable` with `selectableRows`, a controlled `selectedRows`, and `onSelectedRowsChange` writing back to that state.
2. Set `selectedRows` to `[rowA]` from outside the table.
3. Toggle `rowB`'s checkbox → selection should be `[rowA, rowB]` (was `[rowB]` before this fix).
4. Or run the added regression test under `DataTable::controlled selectedRows`.

### DIFF
--- a/src/__tests__/DataTable.test.tsx
+++ b/src/__tests__/DataTable.test.tsx
@@ -255,6 +255,42 @@ describe('DataTable::onSelectedRowsChange', () => {
 	});
 });
 
+describe('DataTable::controlled selectedRows', () => {
+	test('should call onSelectedRowsChange with the correct values when a row is toggled while another is selected via the controlled selectedRows prop', () => {
+		const mock = dataMock();
+		const onSelectedRowsChange = vi.fn();
+
+		const { container, rerender } = render(
+			<DataTable
+				data={mock.data}
+				columns={mock.columns}
+				keyField="id"
+				selectableRows
+				selectedRows={[]}
+				onSelectedRowsChange={onSelectedRowsChange}
+			/>,
+		);
+
+		rerender(
+			<DataTable
+				data={mock.data}
+				columns={mock.columns}
+				keyField="id"
+				selectableRows
+				selectedRows={[mock.data[0]]}
+				onSelectedRowsChange={onSelectedRowsChange}
+			/>,
+		);
+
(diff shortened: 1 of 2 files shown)


jbetancur:
@noahisdai definitely a bug, thanks for logging this issue. This should get released with 8.8.1

Issue #1377, opened 2026-08-24
Feature request: support a minimum width for resizable columns


kitiho:
When column resizing is enabled, could the component also support a configurable minimum width for each column?

For example, a `minWidth` property (or an equivalent option) could prevent users from shrinking a column below a usable size while still allowing it to be resized.

This would help keep table content readable and avoid columns becoming too narrow accidentally.

Thank you for considering it!

kitiho:
Closing this issue for now.

PR #1378, opened 2026-08-24
fix(resize): enforce configured column minWidth

kitiho:
## Bug fix

The resizable-column documentation states that `column.minWidth` clamps how small a user can drag a column. However, the resize handler currently only applies its fixed 40 px floor and ignores the configured `minWidth`. As a result, a column configured with `minWidth: "120px"` can still be dragged to 40 px.

This PR makes the implementation match the documented behavior.

## Changes

- clamp resize drags to an explicitly configured `column.minWidth`
- retain the existing 40 px hard minimum when no column minimum is configured
- add a regression test for dragging below a configured minimum width

## Verification

- `npm test`
- `npm run typecheck`
- `npm run lint`
- `npm run build`

Relates to #1377

### DIFF
--- a/src/__tests__/DataTable.test.tsx
+++ b/src/__tests__/DataTable.test.tsx
@@ -2958,6 +2958,54 @@ describe('DataTable::columnResize', () => {
 		expect(headerCell.style.maxWidth).toBe('60px');
 	});
 
+	test('does not resize a column below its minWidth', () => {
+		const mock = dataMock({ minWidth: '120px' });
+		const { container } = render(<DataTable data={mock.data} columns={mock.columns} resizable />);
+		const handle = container.querySelector('.rdt_resizeHandle') as HTMLElement;
+		const headerCell = handle.closest('[data-column-id]') as HTMLElement;
+
+		Object.defineProperty(headerCell, 'offsetWidth', { value: 300, configurable: true });
+
+		// 300 - 300 = 0, below the configured 120px minimum.
+		fireEvent.pointerDown(handle, { clientX: 300 });
+		fireEvent.pointerMove(document, { clientX: 0 });
+		fireEvent.pointerUp(document);
+
+		expect(headerCell.style.maxWidth).toBe('120px');
+	});
+
+	test('resizes to the 40px floor without a configured minWidth', () => {
+		const mock = dataMock();
+		const { container } = render(<DataTable data={mock.data} columns={mock.columns} resizable />);
+		const handle = container.querySelector('.rdt_resizeHandle') as HTMLElement;
+		const headerCell = handle.closest('[data-column-id]') as HTMLElement;
+
+		Object.defineProperty(headerCell, 'offsetWidth', { value: 300, configurable: true });
+
+		// 300 - 300 = 0, so the existing 40px floor applies.
+		fireEvent.pointerDown(handle, { clientX: 300 });
+		fireEvent.pointerMove(document, { clientX: 0 });
+		fireEvent.pointerUp(document);
+
+		expect(headerCell.style.maxWidth).toBe('40px');
+	});
+
+	test('does not clamp a resize above its minWidth', () => {
+		const mock = dataMock({ minWidth: '120px' });
+		const { container } = render(<DataTable data={mock.data} columns={mock.columns} resizable />);
+		const handle = container.querySelector('.rdt_resizeHandle') as HTMLElement;
+		const headerCell = handle.closest('[data-column-id]') as HTMLElement;
+
+		Object.defineProperty(headerCell, 'offsetWidth', { value: 300, configurable: true });
+
+		// 300 - 50 = 250, above the configured minimum.
+		fireEvent.pointerDown(handle, { clientX: 300 });
+		fireEvent.pointerMove(document, { clientX: 250 });
+		fireEvent.pointerUp(document);
+
+		expect(headerCell.style.maxWidth).toBe('250px');
+	});
+
 	test('inverts drag delta in RTL so dragging the handle left widens the column', () => {
 		const mock = dataMock();
 		const { container } = render(
--- a/src/components/TableCol.tsx
+++ b/src/components/TableCol.tsx
@@ -29,7 +29,7 @@ type TableColProps<T> = {
 	columnDrag: ColumnDragSlice;
 	/** Width override from column resize — takes precedence over column.width */
 	resizedWidth?: number;
-	onResizeStart?: (columnId: string | number, e: React.PointerEvent) => void;
+	onResizeStart?: (columnId: string | number, e: React.PointerEvent, configuredMinWidth?: string) => void;
 	pinnedOffsets?: PinnedOffsets;
 	/** CSS grid placement styles — injected by DataTableHead when rendering in grouped-header grid mode */
 	gridStyle?: React.CSSProperties;
@@ -294,7 +294,7 @@ function TableCol<T>({
 					onPointerDown={e => {
 						// Keep a resize gesture from also starting a column reorder on the cell.
 						e.stopPropagation();
-						onResizeStart(column.id!, e);
+						onResizeStart(column.id!, e, column.minWidth);
 					}}
 					aria-hidden="true"
 				/>
--- a/src/hooks/useColumnResize.ts
+++ b/src/hooks/useColumnResize.ts
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import useIsomorphicLayoutEffect from './useIsomorphicLayoutEffect';
 
-type ResizeRef = { columnId: string | number; startX: number; startWidth: number };
+type ResizeRef = { columnId: string | number; startX: number; startWidth: number; minWidth: number };
 
 export type UseColumnResizeOptions = {
 	initialColumnWidths?: Record<string | number, number>;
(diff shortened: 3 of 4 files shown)


kitiho:
Closing this pull request for now.

src/components/TableCol.tsx
```diff
@@ -29,7 +29,7 @@ type TableColProps<T> = {
 	columnDrag: ColumnDragSlice;
 	/** Width override from column resize — takes precedence over column.width */
 	resizedWidth?: number;
-	onResizeStart?: (columnId: string | number, e: React.PointerEvent) => void;
+	onResizeStart?: (columnId: string | number, e: React.PointerEvent, configuredMinWidth?: string) => void;
```
jbetancur, line 32: i don't think you will need to edit TableCol at all see my comment in the hook

jbetancur:
Hey @kitiho good catch on the bug. Definitely a bug.

The fix doesn't hold up though when I verified this in the browser. 

I think this would be the simpler change, and the tests would need updating for
that scenario:

1. Parse `configuredMinWidth` directly instead of going through the DOM

   ```ts
   const configuredMin = configuredMinWidth != null ? Number.parseFloat(configuredMinWidth) : Number.NaN;
   resizeRef.current = {
       columnId,
       startX: e.clientX,
       startWidth,
       minWidth: Math.max(40, Number.isFinite(configuredMin) ? configuredMin : 0),
   };

I see you commented that you were closing this but it's still open, so if you
want to take a stab at it I'll check back. Otherwise I'll fix this this week.

kitiho:
Updated — thanks for the review.

- removed all changes to `TableCol.tsx`
- the resize hook now looks up the configured column `minWidth` directly, parses it with `Number.parseFloat`, and keeps the existing 40px fallback
- updated the regression test for a 120px configured minimum

Verified with `npm test` (645 passing), `npm run typecheck`, `npm run lint`, and `npm run build`.

kitiho:
## Documentation cleanup

I also removed the duplicate “Drag the right edge of any column header to resize it.” helper text from the Resizable Demo. The `Demo` wrapper already renders this description, so the example now shows it once.

image

src/__tests__/DataTable.test.tsx
```diff
@@ -2958,6 +2958,20 @@ describe('DataTable::columnResize', () => {
 		expect(headerCell.style.maxWidth).toBe('60px');
 	});
```
jbetancur, line 2960: I think we still need a couple more tests.

- resize down to the 40px floor when no minWidth is configured. something like

```ts
	Object.defineProperty(headerCell, 'offsetWidth', { value: 300, configurable: true });

		// 300 - 200 = 100, below the configured 120px minimum.
		fireEvent.pointerDown(handle, { clientX: 300 });
		fireEvent.pointerMove(document, { clientX: 100 });
		fireEvent.pointerUp(document);

		expect(headerCell.style.maxWidth).toBe('120px');
```

- that it does not  clamp a resize that stays above the configured minWidth

```ts
		// Dragging left (clientX 100 → 40) is a -60 delta; in RTL it inverts to +60.
		// startWidth is 0 in jsdom → newWidth = max(40, 0 + 60) = 60
		fireEvent.pointerDown(handle, { clientX: 100 });
		fireEvent.pointerMove(document, { clientX: 40 });
		fireEvent.pointerUp(document);

		expect(headerCell.style.maxWidth).toBe('60px');
```

src/__tests__/DataTable.test.tsx
```diff
@@ -2958,6 +2958,20 @@ describe('DataTable::columnResize', () => {
...
 
+	test('does not resize a column below its minWidth', () => {
+		const mock = dataMock({ minWidth: '120px' });
+		const { container } = render(<DataTable data={mock.data} columns={mock.columns} resizable />);
+		const handle = container.querySelector('.rdt_resizeHandle') as HTMLElement;
+		const headerCell = handle.closest('[data-column-id]') as HTMLElement;
+
```
jbetancur, line 2966: This test passes whether the feature works or not. I think jsdom always reports offsetWidth as 0, so every drag starts from a 0.

Maybe something like this so we start from a real width?

```ts
Object.defineProperty(headerCell, 'offsetWidth', { value: 300, configurable: true });
```

src/components/DataTable.tsx
```diff
@@ -213,6 +208,16 @@ function DataTableInner<T>(props: TableProps<T>, ref: React.ForwardedRef<DataTab
 		defaultSortAsc,
 	);
 
+	// ── Column resize state ────────────────────────────────────────────────────
```
jbetancur: Do we need to move this?

src/hooks/useColumnResize.ts
```diff
@@ -1,42 +1,56 @@
...
 	const [columnWidths, setColumnWidths] = React.useState<Record<string | number, number>>(initialColumnWidths ?? {});
 	const resizeRef = React.useRef<ResizeRef | null>(null);
 	const onColumnResizeRef = React.useRef(onColumnResize);
 	const isRTLRef = React.useRef(isRTL);
+	const columnMinimumsRef = React.useRef(new Map<string | number, string | undefined>());
 	useIsomorphicLayoutEffect(() => {
 		onColumnResizeRef.current = onColumnResize;
 		isRTLRef.current = isRTL;
 	});
+	useIsomorphicLayoutEffect(() => {
```
jbetancur: useLayoutEffect gains nothing over useEffect here, but more importantly the effect is unnecessary

src/hooks/useColumnResize.ts
```diff
@@ -1,42 +1,56 @@
...
+		columnMinimumsRef.current = new Map(columns.map(column => [column.id!, column.minWidth]));
+	}, [columns]);
 
 	const handleResizeStart = React.useCallback((columnId: string | number, e: React.PointerEvent) => {
 		if (typeof document === 'undefined') return;
 		e.preventDefault();
 		const headerCell = (e.currentTarget as HTMLElement).closest('[data-column-id]') as HTMLElement | null;
 		const startWidth = headerCell?.offsetWidth ?? 100;
-		resizeRef.current = { columnId, startX: e.clientX, startWidth };
```
jbetancur, line 31: no need to move this to an effect

kitiho:
@jbetancur 

Thanks for your review ~

But actually I still think the best approach is to pass column.minWidth from TableCol when a resize gesture starts, and parse that value directly in useColumnResize.

For example:
```tsx
// TableCol.tsx
onResizeStart(column.id!, e, column.minWidth);
```

```tsx
// useColumnResize.ts
const configuredMin =
  configuredMinWidth != null ? Number.parseFloat(configuredMinWidth) : Number.NaN;

resizeRef.current = {
  columnId,
  startX: e.clientX,
  startWidth,
  minWidth: Math.max(40, Number.isFinite(configuredMin) ? configuredMin : 0),
};
```
I think this is preferable because `column.minWidth` is already the source of truth for the column’s configured constraint. Passing it at the point where the resize begins keeps the value associated with the specific column being resized, without needing to pass all columns into the hook and build a separate lookup map.
It also avoids reading `minWidth` from the DOM with `getComputedStyle`, which can be less predictable across browser environments and is particularly awkward in JSDOM-based tests.
This keeps the responsibilities clear:
- `TableCol` owns the column configuration and supplies its configured `minWidth`.
- `useColumnResize` owns the resize interaction, parses the configured value, and applies the numeric floor.
- The existing `40px` minimum remains the fallback when no valid `minWidth` is configured.
Compared with maintaining column minimums in a ref through an effect, this is a more direct data flow and removes the extra `columns` option, `Map`, ref, and effect from `useColumnResize`.

jbetancur:
Fair point on TableCol @kitiho, let's go with your version. Passing `column.minWidth` at the
start of the gesture is cleaner and it drops the columns option,
the map, the ref and the effect. That also means the useColumnResize call in
DataTable.tsx can move back where it was.

Ignore my two comments on the layout effect then, that code goes away with this.

Tests still need work though. Sorry, I swapped the snippets in my earlier comment. The
40px floor one should use a column with no minWidth:

```tsx
const mock = dataMock();
Object.defineProperty(headerCell, 'offsetWidth', { value: 300, configurable: true });

// 300 - 300 = 0, so the 40px floor applies.
fireEvent.pointerDown(handle, { clientX: 300 });
fireEvent.pointerMove(document, { clientX: 0 });
fireEvent.pointerUp(document);

expect(headerCell.style.maxWidth).toBe('40px');

// And the one that shouldn't clamp:


const mock = dataMock({ minWidth: '120px' });
Object.defineProperty(headerCell, 'offsetWidth', { value: 300, configurable: true });

// 300 - 50 = 250, above the minimum, left alone.
fireEvent.pointerDown(handle, { clientX: 300 });
fireEvent.pointerMove(document, { clientX: 250 });
fireEvent.pointerUp(document);

expect(headerCell.style.maxWidth).toBe('250px');

kitiho:
Thanks for the detailed review and for clarifying the test cases. @jbetancur 

I updated the implementation to pass `column.minWidth` from `TableCol` when the resize gesture starts, then parse and clamp that value directly in `useColumnResize`. This removes the `columns` option, the minimum-width map/ref/effect, and lets the resize setup in `DataTable` return to its original position.

I also updated the tests to cover:

- clamping to a configured `minWidth`
- preserving the existing `40px` floor when no `minWidth` is configured
- allowing resize values that remain above the configured minimum

All checks pass locally: `npm test` (647 passing), `npm run typecheck`, `npm run lint`, and `npm run build`.

Thanks again for the guidance — it made the implementation much simpler and more focused.

jbetancur:
Thank you @kitiho 8.8.2 should be available shortly!

PR #1387, opened 2026-09-05
fix: keep column filter popup within the viewport

### LINKED ISSUE #1386: [BUG]: Column filter popup can clip off screen
### Checklist

- [x] I agree to the [Code of Conduct](https://github.com/jbetancur/react-data-table-component/blob/master/CODE-OF-CONDUCT.md)
- [x] I searched [existing issues](https://github.com/jbetancur/react-data-table-component/issues?q=is%3Aissue) and this is not a duplicate
- [x] I read the [documentation](https://reactdatatable.com)

### Describe the bug

The column filter selector can clip out of the viewport making it unusable

Image

### Steps to reproduce

1. Create a very long table with filter column
2. Click the filter column button
3. See above.

### Expected behavior

The control should likely stay with in X pixels within the viewport

### Actual behavior

It clips out.

### Minimal reproduction

n/a

### react-data-table-component version

8.9.0

### React version

19.8.2

### Browser & OS

Firefox, MacOS latest

dajiaohuang:
## Summary

Fixes #1386.

The filter popup is positioned using a hardcoded 260px width and always opens below its button. Its rendered width can be larger (for example, datetime inputs), and a button near the bottom of the viewport leaves the controls off screen.

Measure the rendered panel before paint, using the existing isomorphic layout-effect pattern from ContextMenu. Clamp it with an 8px viewport margin, prefer opening below the button, then flip above or fit within the viewport when necessary. Cap its dimensions and allow internal scrolling in very small viewports. Remeasure after parent renders so adding a condition stays within the viewport.

No public API or dependency changes. Existing dismissal and focus behavior is preserved.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation / example update

## Checklist

- [x] I read [CONTRIBUTING.md](https://github.com/jbetancur/react-data-table-component/blob/master/CONTRIBUTING.md)
- [x] I agree to the [Code of Conduct](https://github.com/jbetancur/react-data-table-component/blob/master/CODE-OF-CONDUCT.md)
- [x] Tests pass (`npm test`)
- [x] I have updated docs or examples if needed (not needed: existing behavior is restored; no API changes)

## How to test

- `npm run lint`, `npm run typecheck`, `npm test`: pass, including six new positioning regressions (692 tests total). Four of the new tests fail against the original implementation.
- `npm run build` and `npm run docs:build`: pass. On Windows the existing Git-for-Windows `rm.exe` was made available on the process PATH for the library build.
- Render a filter button near the bottom-right viewport edge, open its popup, and add a second condition. It should stay visible. Repeat with text, number, date, datetime, time, and set filters in a short/narrow viewport; overflowing content should scroll inside the panel.

Additional local headless Edge 152 checks used the actual component and stylesheet, with no browser tooling added to this repository. Across 800x600, 320x240, and 220x120 viewports, all six filter types remained within the 8px boundary, including condition growth and set-list search/reset. Apply, Escape, initial focus, focus return, and resize dismissal passed. At 800x600 the original datetime popup bounds were `(532, 576, 288, 103)` (clipped right/bottom); the fix produced `(504, 443, 288, 103)`.

The reported Firefox/macOS environment was not available locally; the browser reproduction above is Edge on Windows with the repository's React 18 dependency.

### DIFF
--- a/src/DataTable.css
+++ b/src/DataTable.css
@@ -452,8 +452,10 @@
 	border-radius: var(--rdt-border-radius, 6px);
 	box-shadow: 0 4px 16px rgba(0, 0, 0, 0.14);
 	padding: 10px;
-	min-width: 260px;
-	max-width: calc(100vw - 24px);
+	min-width: min(260px, calc(100vw - 16px));
+	max-width: calc(100vw - 16px);
+	max-height: calc(100vh - 16px);
+	overflow: auto;
 	box-sizing: border-box;
 }
 
--- a/src/__tests__/ColumnFilter.position.test.tsx
+++ b/src/__tests__/ColumnFilter.position.test.tsx
@@ -1,0 +1,79 @@
+import * as React from 'react';
+import { render, fireEvent } from '@testing-library/react';
+import ColumnFilter from '../components/ColumnFilter';
+import { emptyFilterState } from '../hooks/useColumnFilter';
+
+describe('ColumnFilter:viewport positioning', () => {
+	let anchor: DOMRect;
+	let panelWidth: number;
+	let panelHeight: number;
+
+	beforeEach(() => {
+		vi.stubGlobal('innerWidth', 800);
+		vi.stubGlobal('innerHeight', 600);
+		anchor = new DOMRect(40, 40, 22, 22);
+		panelWidth = 340;
+		panelHeight = 160;
+		vi.spyOn(HTMLElement.prototype, 'getBoundingClientRect').mockImplementation(function (this: HTMLElement) {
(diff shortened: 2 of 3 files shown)


jbetancur:
@dajiaohuang thank you for the PR and catching this one. Once thing to note is that this does appear to handle RTL cases. I will open a separate issue to handle RTL use cases.

jbetancur:
I'll release an 8.9.1 shortly

jbetancur:
#1388 opened to address RTL issue in future

PR #1393, opened 2026-09-08
feat(docs): synchronize site and demo color modes

### LINKED ISSUE #1381: Dark mode for the docs site
The library supports dark mode. The docs site does not. `colorMode`, `useColorMode`, dark theme overrides and tests all ship in `src/`, but reactdatatable.com is light only, so there is no way to see any of it.

Three layers:

**1. Code blocks.** Astro/Shiki has a config setting for this. Currently single-theme:

```js
shikiConfig: { theme: "catppuccin-macchiato" }
```

Switching to `themes: { light, dark }` emits both palettes as CSS variables (`--shiki-dark`). Astro does not switch between them, we add the CSS. Two things conflict today: `global.css` hardcodes `bg-gray-950` over code blocks, and `Demo.astro` / `CodeBlock.astro` both pass `theme="catppuccin-macchiato" defaultColor={false}` straight to `<Code>`, bypassing the config.

**2. Site chrome.** Body, sidebar, prose, tables, prop tables. Hand-written `dark:` variants. Tailwind is already installed, nothing in `global.css` uses it yet, and `body` is hardcoded `bg-white text-gray-900`.

**3. Demos.** Already handled by the library, just not wired up. `LiveDemo.tsx` hardcodes `colorMode="light"`. `ThemedDataTable` is a pure passthrough today even though it exists to be the seam for this.

**Toggle contract is already fixed by the library.** `useColorMode` reads `localStorage.theme`, then `html.dark`, then `prefers-color-scheme`, and watches all three. So a header toggle writing `localStorage.theme` and toggling `html.dark` makes every demo follow along with no library change, and `ThemedDataTable` can default `colorMode` to `system`.

Needs an inline script in `Layout.astro` head to set the class before first paint, otherwise dark loads with a white flash.

Note `ThemesDemo.tsx` already has `dark:` classes and a label reading "follows the page toggle above", pointing at a toggle that was never built.

dajiaohuang:
## Summary

Fixes #1381. Adds a Light / Dark / System selector to the documentation header, initializes the page theme before first paint, and synchronizes the site, code blocks, and demos through the existing `localStorage.theme` contract. System mode follows OS changes; explicit preferences persist across navigation and synchronize between tabs.

The browser check also exposed a hydration bug in `useColorMode`: a dark client could retain the server's light inline styles because its initial state was already dark. Initialize consistently with the server, then apply the detected preference in the effect. A regression test fails before this change and passes afterward. Demo wrappers default to system mode while preserving explicit demo overrides.

## Type of change

- [x] Bug fix
- [x] New feature
- [x] Breaking change
- [x] Documentation / example update

## Checklist

- [x] I read [CONTRIBUTING.md](https://github.com/jbetancur/react-data-table-component/blob/master/CONTRIBUTING.md)
- [x] I agree to the [Code of Conduct](https://github.com/jbetancur/react-data-table-component/blob/master/CODE-OF-CONDUCT.md)
- [x] Tests pass (`npm test`)
- [x] I have updated docs or examples if needed

## How to test

1. Run `npm run docs:build`, then preview the built site.
2. Open `/docs/themes/` with a dark OS preference and no saved theme. The page and hydrated tables should both be dark.
3. Select Light, reload, then return to System and change the OS preference. Check the table and code-block palettes along with the page.
4. Change the selection in another tab, visit the homepage and a recipe demo, and use the selector at a 390px viewport.

Validation: 719 tests passed; lint and typecheck passed; library build passed using Git Bash for the existing POSIX build command on Windows; documentation build produced 55 pages. Browser acceptance covered 14 checks including first hydration, persistence, OS changes, cross-tab synchronization, homepage/recipe/headless demos, and mobile controls, with no JavaScript runtime errors.

The mobile check covers the new control and theme behavior. Wide table/reference content is not claimed to be overflow-free.


jbetancur:
Thanks @dajiaohuang It's been a busy week but I will  have a look this weekend

jbetancur:
This is great @dajiaohuang! However, I think it still needs some polishing.

http://localhost:4399/docs/conditional-styles
the table cels are washed out

http://localhost:4399/docs/selection
info section is a bit washed out

Also there is a strange flashing of tables when a page is loaded in dark mode

There are also several other demo controls that might not be inheriting the theme.

dajiaohuang:
Follow-up pushed in a010b82.

- Conditional-styles rows/cells now use dark-aware CSS variables, including readable muted values and status backgrounds.
- Selection feedback and controls have dark-mode contrast states.
- Demo surfaces remap remaining light Tailwind utilities in dark mode.
- System color-mode resolution now uses an isomorphic layout effect, so hydrated tables apply the pre-paint html.dark marker before the first browser paint.

Validation: npm test (719 tests), npm run typecheck, npm run build, npm run docs:build, and targeted Prettier checks pass. The docs build reports the existing Pagefind note that /docs/ has no outer html element. Remote CI/deploy checks will run against a010b82.

PR #1394, opened 2026-09-14
fix(a11y): point pinned scrollbar aria-controls at the scroll container's actual id

askalf:
## Summary

- `PinnedScrollbar` used this mount's `React.useId()` for `aria-controls`, but assigned that id to the scroll container only when `el.id` was empty.
- A pre-existing container id—whether supplied by the host, retained through a remount, or reached after a `scrollRef` swap—therefore left the thumb pointing at a nonexistent id.
- The fix stores the actual `el.id` after the existing stamp-or-preserve branch and renders `aria-controls={controlsId}`. It changes 10 source lines and no public API.

```console
$ npx vitest run src/__tests__/pinning.test.tsx     # fixed source
 ✓ src/__tests__/pinning.test.tsx (24 tests) 888ms
 Test Files  1 passed (1)
      Tests  24 passed (24)
   Duration  3.86s

$ git checkout master -- src/components/PinnedScrollbar.tsx
$ npx vitest run src/__tests__/pinning.test.tsx     # base source; identical 24-test file
 × pinned scrollbar aria-controls still resolves after a column is unpinned and re-pinned
 × aria-controls points at a host-supplied container id instead of a fresh one
 × aria-controls survives a remount onto the same container
 × aria-controls re-points when the scroll container is swapped for another one
 × aria-controls follows an id the host assigns after the first mount stamped one
 ✓ aria-controls resolves to the scroll container it labelled (control)
 ✓ two concurrent scrollbars each control their own container (control)
 ✓ emits no thumb and no aria-controls when the scroll ref is empty (control)
 Test Files  1 failed (1)
      Tests  5 failed | 19 passed (24)
   Duration  3.29s
```

## Decisions

`PinnedScrollbar` now keeps a `controlsId` state initialized from `scrollContainerId`; in the existing synchronization effect, `setControlsId(el.id)` is called immediately after the existing `if (!el.id) { el.id = scrollContainerId; }` stamp-or-preserve branch, and `aria-controls={controlsId}` is rendered instead of the raw `scrollContainerId`. That way the element's actual id is used whether it was freshly stamped or already present.

Rejected alternatives: mutating a ref directly wouldn't trigger a re-render of the thumb; overwriting `el.id` unconditionally would clobber host-owned DOM attributes; adding a `MutationObserver` to track id changes during an already-active mount is a different, unrelated concern and is out of scope for this fix.

Full package suite (`npm test`, 23 files / 726 tests) and `npx tsc --noEmit` both pass with the fix; not run: any real-browser/layout-specific check, since the defect and its tests are DOM id/reference resolution, fully exercised under jsdom.

AI assistance: this bug was found and the fix and tests were drafted with AI tooling in my workflow; the tests and checks above were executed as pasted. I'm responsible for the change and will handle review feedback.


src/__tests__/pinning.test.tsx
```diff
@@ -195,4 +221,185 @@ describe('PinnedScrollbar', () => {
...
+		const thumb = second.container.querySelector('.rdt_pinnedScrollbarThumb') as HTMLElement;
+		expect(thumb.getAttribute('aria-controls')).toBe('host-took-over-later');
+		expect(document.getElementById(thumb.getAttribute('aria-controls')!)).toBe(ref.current);
+
+		second.unmount();
+		ref.current!.remove();
+	});
+
+	// Control: two scrollbars on the page at once, each on its own container.
+	// Green before and after — the id each thumb points at is per-instance state,
```
jbetancur: I know this is what the AI generated but can we clean this comment up. Like Green before and after. just indiciate what the tests accomplishes

src/__tests__/pinning.test.tsx
```diff
@@ -195,4 +221,185 @@ describe('PinnedScrollbar', () => {
 			expect(ref.current!.scrollLeft).toBeGreaterThanOrEqual(0);
 		}
 	});
+
+	// Control: the container has no id of its own, so the effect stamps it with
+	// this mount's useId and aria-controls matches either way. Green before and
```
jbetancur: I know this is what the AI generated but can we clean this comment up. Like Green before and after. just indiciate what the tests accomplishes

src/components/PinnedScrollbar.tsx
```diff
@@ -13,6 +13,9 @@ export default function PinnedScrollbar({
 	rightInset,
 }: PinnedScrollbarProps): JSX.Element | null {
 	const scrollContainerId = React.useId();
+	// The id the thumb points at. Only equals scrollContainerId when this
```
jbetancur: This is kind of a nit and won't block the PR -  and I should put it in the claude.md but try to remove em dashes.

jbetancur (changes requested):
@askalf Thank you for identifying this bug! Just a few comments to address. Also please update the CHANGELOG.md. One generate nit is to avoid those em dashes if you can. Not a blocker and I will add this to the claude.md in the future

askalf:
Thanks for the review. All three addressed in the latest push:

- The two control-test comments now just say what each test pins and that it is green before and after the fix.
- Every em dash this change introduced is gone, tests and component both. Happy to keep that convention for anything else here.
- CHANGELOG entry added under 8.10.0 / Bug fixes.

`pinning.test.tsx` passes locally on the new head, 24/24.

src/__tests__/pinning.test.tsx
askalf: Done in 9913338: the comments now say what each test checks, nothing about how it was written.

src/__tests__/pinning.test.tsx
askalf: Done in 9913338: the comments now say what each test checks, nothing about how it was written.

src/components/PinnedScrollbar.tsx
askalf: Done in 9913338, no em dashes left in the diff. CHANGELOG entry is in the same push.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ A summary of notable changes per release. For the full commit history see the [r
 
 ### Bug fixes
 
+- **Pinned scrollbar** now sets `aria-controls` to the scroll container's real id. When the container already carried an id, or was labelled again after an unpin and re-pin, the thumb kept pointing at a generated id no element had, so assistive technology could not resolve what the scrollbar controlled.
```
jbetancur, line 21: Sorry I should have been more clear. This needs to land in a new entry for 8.10.1 since 8.10.0 was already released

askalf:
Updated the test-only cleanup after the CI report: the `for-of` cleanup now has braces to satisfy the repository curly rule, and `attach` preserves its concrete element type so the swap harness remains type-safe. The full CI chain is clean locally: ESLint, `tsc --noEmit`, Prettier, the focused 24-test Vitest file, and the production build. The current upstream run is awaiting first-contributor approval.

